PR #61, opened 2019-05-08
Adding auto start settings SAL-608

dantrinh:
The new autostart settings page is a static component at the moment. It is located in components/ of the react project! It uses settings.css and is a stories.tsx file at the moment


dsarfati:
@thedigitaldesign Is going to take this over. Closing for now.

PR #70, opened 2019-05-29
SAL-669: Make selected reward clickable

thedigitaldesign:
Added an onClick to show the modal of the selected reward

### DIFF
--- a/packages/web-app/src/modules/reward-views/components/SelectedReward.tsx
+++ b/packages/web-app/src/modules/reward-views/components/SelectedReward.tsx
@@ -28,6 +28,7 @@ const styles = (theme: SaladTheme) => ({
     height: '100px',
     position: 'relative',
     justifyContent: 'flex-start',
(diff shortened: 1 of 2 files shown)


packages/web-app/src/modules/reward-views/components/SelectedReward.tsx
```diff
@@ -110,6 +111,14 @@ class _SelectedReward extends Component<Props> {
     return ' '
   }
 
+  handleClick = (reward: Reward) => {
```
dsarfati: Maybe instead of passing in the reward, you could simply pull the reward out of `this.props` on Ln 115. This will eliminate the need for the arrow function on Ln 130. Arrow functions like this get created each time a component is rendered.

dsarfati:
Looking great! We also need to change the cursor to the `pointer` for the selected reward so the user knows that it is clickable.

PR #73, opened 2019-06-12
SAL-647: Update cuda driver error with ethminers new code

thedigitaldesign:
Added in the check for the error and changed the if/else into a switch where the cuda errors will fall through to the shared modal

### DIFF
--- a/packages/desktop-app/src/config.ts
+++ b/packages/desktop-app/src/config.ts
@@ -48,8 +48,8 @@ export const convertMinutes = (hours: number): number => hours * 60000
 
 class Config {
   public readonly devTools: boolean = false
-  public readonly appUrl: string = 'http://localhost:3000/'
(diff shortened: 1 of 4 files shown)


packages/desktop-app/src/config.ts
```diff
@@ -48,8 +48,8 @@ export const convertMinutes = (hours: number): number => hours * 60000
 
 class Config {
   public readonly devTools: boolean = false
-  // public readonly appUrl: string = 'http://localhost:3000/'
-  public readonly appUrl: string = 'https://app.salad.io'
+  public readonly appUrl: string = 'http://localhost:3000/'
```
dsarfati: :) This really needs to get moved into a config file :)

thedigitaldesign: I was thinking the same thing :). Could we just gitignore it?

PR #79, opened 2019-06-28
Feature/DEV-54 and DEV-40

thedigitaldesign:
- A lot of rewriting of AuthStore
- Fixes to logic in Routes and App
- Switched Auth0 to Salad token and expiration


packages/web-app/src/App.tsx
```diff
@@ -1,112 +1,41 @@
...
 
-  getOnboardingRedirect = () => {
-    let profile = this.store.profile.currentProfile
-    if (profile === undefined) {
-      if (this.store.profile.isLoading) return <Redirect to="/profile-loading" />
-      else return null
+  componentDidMount = () => {
+    if (this.store.auth.isAuth) {
+      this.store.profile.loadProfile()
+        .then(() => { })
```
dsarfati: Were you going to do anything in the then or was this just from your testing?

packages/web-app/src/modules/auth/AuthStore.ts
```diff
@@ -51,56 +63,68 @@ export class AuthStore {
...
-
-      yield this.store.profile.loadProfile()
-
-      this.loginError = false
-      this.isLoading = false
-      this.store.routing.push('/')
+      yield this.webAuth.parseHash((err, authResult) => {
+        if (authResult) {
+          this.axios.post('generate-salad-token', { authToken: authResult.accessToken })
+            .then((response) => {
```
dsarfati: Since this is within an async method (via MobX `flow`) these can use `yield` instead of doing `.then`.

packages/web-app/src/modules/auth/AuthStore.ts
```diff
@@ -51,56 +63,68 @@ export class AuthStore {
...
+    this.authToken = token.saladToken
+    this.expiresAt = token.expires
+    this.axios.defaults.headers.common['Authorization'] = `Bearer ${token.saladToken}`
+
+    if (token.saladToken) {
+      this.isAuth = true
+      this.loginError = false
+      this.isLoading = false
+
+      // Uncaught Error: Maximum update depth exceeded.
```
dsarfati: Remove?

packages/web-app/src/modules/machine/NativeStore.ts
```diff
@@ -108,8 +108,8 @@ export class NativeStore {
 
         store.analytics.captureException(new Error(`Received error code ${errorCode} from native`))
 
-        switch(errorCode) {
-          case 8675309: 
+        switch (errorCode) {
```
dsarfati: I have noticed a few other whitespace changes. I have a prettier file that should keep these all consistent. Check to see if you are using that config, it's possible this file never was updated with our standard formatting.

packages/web-app/src/modules/profile/ProfileStore.ts
```diff
@@ -33,25 +36,33 @@ export class ProfileStore {
...
-        this.currentProfile.referred === ReferredStatus.CanEnter)
+    const onboarding = (
+      this.currentProfile === undefined
+      || (this.currentProfile.termsOfService !== Config.termsVersion
+        || this.currentProfile.whatsNewVersion !== Config.whatsNewVersion
+        || this.needsAnalyticsOnboarding
+        || this.currentProfile.referred === ReferredStatus.CanEnter)
     )
+
+    this.setOnboarding(onboarding)
```
dsarfati: I'm not sure if there are any issues with side-effects within `@computed` functions. Is there a reason for keeping both the `@computed` and the property `this.onboarding`?

packages/web-app/src/modules/auth/AuthStore.ts
thedigitaldesign: When I try using yield within the webAuth yield, I get an error, 

`A 'yield' expression is only allowed in a generator body`

packages/web-app/src/modules/machine/NativeStore.ts
thedigitaldesign: I was using a different extension, I'll get Prettier in place. I was just hitting Alt + Shift + F to format the file.

packages/web-app/src/modules/profile/ProfileStore.ts
thedigitaldesign: This is a artifact from trying to figure out issues with the blue screen bug, I noticed we were calling isOnboarding and an authorization function multiple times. I was trying to set it once and use the observable. I don't think I need this logic any more, will revert.

PR #80, opened 2019-07-02
Added Canny

dsorensen00:
Canny and cool carrot logo


dsarfati:
@dsorensen00 We need to connect the bug report button on the settings page to also take us to this page. I'm going to WIP this until we have better product requirements and know exactly what the roll out looks like for this tool.

dsarfati:
After talking with @mkjenkins today i think the support button at the bottom of the app should go to `salad.io/support` and the Bug Report button should go to our Canny page.

PR #86, opened 2019-07-17
feature/DEV-42_windows-autostart

thedigitaldesign:


packages/web-app/src/modules/settings-views/components/Settings.tsx
```diff
@@ -37,6 +40,10 @@ interface Props extends WithStyles<typeof styles> {
 class _Settings extends Component<Props> {
   store = getStore()
 
+  componentDidMount() {
+    document.addEventListener('keyup', this.handleCloseKeyPress)
```
dsarfati: Do we need to clean up the event listener and remove it when the component unmounts?

PR #88, opened 2019-07-22
feature/DEV-190_v0.2.2-Login

thedigitaldesign:
- Overriding the axios baseURL
- Sending new payload

### DIFF
--- a/packages/web-app/src/modules/auth/AuthStore.ts
+++ b/packages/web-app/src/modules/auth/AuthStore.ts
@@ -90,6 +90,7 @@ export class AuthStore {
                 systemId: this.store.native.machineInfo
                   ? this.store.native.machineInfo.system.uuid
                   : 'ffd15a2b-ee3a-498e-9975-389d7d46161d',
+                idToken: authResult.idToken,
               },
               {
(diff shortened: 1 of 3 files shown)


packages/web-app/src/modules/auth/AuthStore.ts
```diff
@@ -64,22 +64,37 @@ export class AuthStore {
...
-          this.axios.post('generate-salad-token', { authToken: authResult.accessToken })
-            .then((response) => {
+          this.axios
+            .post(
+              'login',
+              {
+                authToken: authResult.accessToken,
+                systemId: this.store.native.machineInfo
+                  ? this.store.native.machineInfo.system.uuid
+                  : 'ffd15a2b-ee3a-498e-9975-389d7d46161d',
```
dsarfati: How do we ensure that we have the machine info before the `handleAuthentication` method is called? Maybe we should move the logic of getting the machine info into this method so it's all inline

thedigitaldesign: It is, but not in this PR. The logic is in the DEV-185 since it was related to that task.

PR #92, opened 2019-08-04
Feature/dev 185 v0.2.1 connect worker status

thedigitaldesign:
- Removed worker status heartbeat
- Changed login and pulled Salad token logic out into it's own class
- Updated machineId and nanopool will now use machineId as the worker


packages/web-app/src/modules/auth/AuthStore.ts
```diff
@@ -2,6 +2,7 @@ import { action, observable, flow } from 'mobx'
 import { AxiosInstance } from 'axios'
 import { WebAuth } from 'auth0-js'
 import { RootStore } from '../../Store'
+import uuidv1 from 'uuid/v1'
```
dsarfati: I think you want to use `v4` instead of `v1` since `v4` is a random guid and `v1` is based on the timestamp. In theory this reduces the risk of duplicate uuids across users

packages/web-app/src/modules/auth/TokenStore.ts
```diff
@@ -0,0 +1,45 @@
...
+import { TokenData } from './models'
+import * as Storage from '../../Storage'
+
+const SALAD_TOKEN = 'TOKEN'
+
+export class TokenStore {
+  @observable
+  public saladToken: string = ''
+
+  @action
```
dsarfati: Don't think this needs to be an `@action` since it doesn't modify any `@observable` properties

packages/web-app/src/modules/auth/TokenStore.ts
```diff
@@ -0,0 +1,45 @@
...
+  getToken = () => Storage.getItem(SALAD_TOKEN)
```
dsarfati: Also, why is this used when the could just use the `.saladToken` property to get the token?

PR #93, opened 2019-08-04
Connected reward list to the to new APIs

dsarfati:


thedigitaldesign:
@dsarfati My only concern with this PR is the commented out code. I don't have an understanding of what we are changing/keeping with the new rewards UI.

dsarfati:
@thedigitaldesign at this point all the commented out code is for reward redemption. Once we get to that we/I will fix all that code and we can remove all the commented out code.

PR #103, opened 2019-08-09
feature/DEV-77_v0.2.1-Hook-up-Onboarding-API

thedigitaldesign:
- Switched over Profile to new API …
- Ran through heavy debugging for a multiple loading of Terms, found the issue and commented out registerMachine()


packages/web-app/src/modules/profile/models/Profile.ts
```diff
@@ -2,12 +2,9 @@ export interface Profile {
   id: string
   username: string
   email: string
-  isNewUser: boolean
-  termsOfService?: string
-  whatsNewVersion?: string
-  referred?: ReferredStatus
-  trackUsageVersion?: string
-  tutorialComplete: boolean
+  lastAcceptedTermsOfService: string | null | undefined
```
dsarfati: Why include `null`?

packages/web-app/src/modules/profile/models/ProfileResource.ts
```diff
@@ -1,14 +1,12 @@
...
-  termsOfService: string
-  trackUsageVersion: string
-  tutorialComplete: number
-  whatsNewVersion: string
-  profileData: {
-    email: string
-    name: string
-    nickname: string
+  id: string
+  auth0Id: string
```
dsarfati: Is the `auth0Id` required here? I don't return this value now

packages/web-app/src/modules/profile/models/ProfileResource.ts
```diff
@@ -1,14 +1,12 @@
...
+  created: string
+  username: string
+  email: string
+  onboarding: {
```
dsarfati: The API resource does not have these values under an `onboarding` property. Just wanna make sure that we are in sync here.

packages/web-app/src/modules/profile/models/Profile.ts
thedigitaldesign: Removed. It was leftover from when I was unclear on how things were setup.

packages/web-app/src/modules/profile/models/ProfileResource.ts
thedigitaldesign: Removed

packages/web-app/src/modules/profile/models/ProfileResource.ts
thedigitaldesign: Flattened

PR #105, opened 2019-08-10
Use load balanced API to eliminate CORS pre-flight requests DEV-217

seniorquico:
@dsarfati and @thedigitaldesign, are there any other configuration values (or hardcoded constants) that need to get updated?

With the new network topology, all traffic to `https://app.salad.io` will get routed to the load balancer. The load balancer will then route traffic as follows (in order):

> 1. https://app.salad.io/api/* -> API
> 2. https://app.salad.io/* -> Application

The `app-api.salad.io` subdomain will be retired.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -18,6 +18,12 @@ ID = "salad-web-app"
   # Default build command.
   command = "yarn build"
 
+# The following redirect proxies API requests to the backend.
+[[redirects]]
+  from = "/api/*"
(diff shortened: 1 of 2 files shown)


dsarfati:
There is one more place in the `.env` file. https://github.com/SaladTechnologies/salad-applications/blob/9d0168331d9c5c21cd68cd0804490c9b267b75b8/packages/web-app/.env#L5 

We will be cleaning up/consolidating these files soon

PR #113, opened 2019-08-14
feature/v021/DEV-314-System-Info-API

thedigitaldesign:


dsarfati:
@thedigitaldesign Here is your build error
```
3:07:55 PM: Type error: Cannot find module '../../machine/models/Machine'.  TS2307
3:07:55 PM:     2 | import { storiesOf } from '@storybook/react'
3:07:55 PM:     3 | import { MachineSummaryView } from './MachineSummaryView'
3:07:55 PM:   > 4 | import { Machine } from '../../machine/models/Machine'
3:07:55 PM:       |                         ^
3:07:55 PM:     5 | import { StartButton } from './StartButton'
3:07:55 PM:     6 | import { action } from '@storybook/addon-actions'
3:07:55 PM:     7 | import { CompatibilityCheckPage } from './CompatibilityCheckPage'
3:07:55 PM: error Command failed with exit code 1.
```

packages/web-app/src/modules/profile/ProfileStore.ts
```diff
@@ -59,9 +59,8 @@ export class ProfileStore {
     try {
       let profile = yield this.axios.get('profile')
       this.currentProfile = profile.data
-
-      // NOTE: Causes multiple loads of T&C. This is being wired up this sprint...
-      // yield this.store.native.registerMachine()
+      
+      yield this.store.native.registerMachine()
```
dsarfati: There is a new `onLogin` method in the `RootStore` that this could move to. That way we can keep machine stuff out of profiles. Let me know what you think about this pattern.

packages/web-app/src/modules/machine/NativeStore.ts
```diff
@@ -279,14 +266,14 @@ export class NativeStore {
     }
 
     try {
-      let res = yield this.axios.post('/check-gpu', req, { baseURL: 'https://api.salad.io/core/master/' })
+      let res = yield this.axios.post('check-gpu', req, { baseURL: 'https://api.salad.io/core/master/' })
```
dsarfati: I think we are going to eliminate this API in the new system by combining it with the register machine API. When you POST the machine data to `/machine/{id}/data` that returns back a `MachineResource` that has a property telling you if it is qualifying or not. I think that makes things a bit simpler.

packages/web-app/src/modules/machine/NativeStore.ts
```diff
@@ -243,27 +240,20 @@ export class NativeStore {
...
-
     try {
       console.log('Registering machine with salad')
-      let res = yield this.axios.post('register-machine', deviceInfo, {
-        baseURL: 'https://api.salad.io/core/master/',
-      })
+      let res: any = yield this.axios.post(`machines/${this.machineId}/data`, this.machineInfo)      
       console.log(res)
+
+      this.validGPUs = res.data.validGpus
```
dsarfati: I think we also need to set the `this.validOperatingSystem` so we could show the correct error page if they are using windows 8

thedigitaldesign: Added

PR #144, opened 2019-09-01
feature/v030/DEV-393_Connect-heartbeat

thedigitaldesign:
[DEV-393](https://saladtech.atlassian.net/browse/DEV-393)
- @dsarfati When mining starts, status is sent but am getting a 409, "The machine is already running." Afterwards, when the miner is still initializing I don't believe I am getting a 409 again.

[DEV-433](https://saladtech.atlassian.net/browse/DEV-433)

[DEV-434](https://saladtech.atlassian.net/browse/DEV-434)
- This is currently set to 5 heartbeats of 0 hashrate reported

There are some things we'll need to talk about when the team is back. The mining status is also on the app.


packages/web-app/src/modules/machine/NativeStore.ts
```diff
@@ -62,14 +71,21 @@ export class NativeStore {
...
+
   @computed
   get isNative(): boolean {
     return window.salad && window.salad.platform === 'electron'
   }
 
   @computed
   get isCompatible(): boolean {
-    return true // TODO: 0.2.1 - this.isNative && this.validOperatingSystem && this.validGPUs
+    // return this.isNative && this.validOperatingSystem && this.validGPUs
```
dsarfati:
```suggestion
    // TODO: return this.isNative && this.validOperatingSystem && this.validGPUs
```

dsarfati:
@thedigitaldesign There are now 3 different statuses that you can send to the API, `Initializing`, `Running`, & `Stopped`
The intention was that you would only send `Initializing` once, then you would keep sending heartbeats with status of `Running` (regardless of the miner status). If you receive a 409, that just means that the status has already changed and that the client and backend are out of sync. It is same to assume that if you send `Initializing` and get a 409 back you can ignore that error and just keep sending the `Running` heartbeats. If you send `Stopped` and get a 409 just make sure that the client is showing as `Stopped`

PR #149, opened 2019-09-04
Updated Alpha Badge DEV-418

dsorensen00:
Alpha Badge copy updated.

### DIFF
--- a/packages/web-app/src/modules/profile-views/components/ProfileMenuTooltip.tsx
+++ b/packages/web-app/src/modules/profile-views/components/ProfileMenuTooltip.tsx
@@ -13,10 +13,11 @@ const styles = (theme: SaladTheme) => ({
     paddingTop: '5px',
     pointerEvents: 'none',
   },
-  contentContiner: {
+  contentContainer: {
     padding: '0 1rem 1rem',
     fontFamily: 'sharpGroteskLight25',
     fontSize: theme.small,
+    marginTop: '0px !important',
   },
 })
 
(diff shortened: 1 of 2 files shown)


packages/web-app/src/modules/profile-views/components/ProfileSummaryTitle.tsx
```diff
@@ -7,6 +7,7 @@ import classnames from 'classnames'
 const styles = (theme: SaladTheme) => ({
   container: {
     padding: '1rem',
+    paddingBottom: '0px !important',
```
dsarfati: Instead of setting all the padding values and then overriding one of the values to 0 using `!important` you should just update the `padding` to use one of the other overloads that sets left, top and right

packages/web-app/src/modules/profile-views/components/ProfileMenuTooltip.tsx
```diff
@@ -13,10 +13,11 @@ const styles = (theme: SaladTheme) => ({
     paddingTop: '5px',
     pointerEvents: 'none',
   },
-  contentContiner: {
+  contentContainer: {
     padding: '0 1rem 1rem',
     fontFamily: 'sharpGroteskLight25',
     fontSize: theme.small,
+    marginTop: '0px !important',
```
dsarfati: Why does this need to be `!important`, that usually is a sign that things should be reorganized

packages/web-app/src/modules/profile-views/components/ProfileMenuTooltip.tsx
```diff
@@ -33,17 +34,11 @@ class _ProfileMenuTooltip extends Component<Props> {
     return (
       <div className={classnames(className, classes.container)}>
         <ProfileSummaryTitle profile={profile} xp={xp} />
-        <div className={classnames(classes.contentContiner, 'content')}>
+        <div className={classnames(classes.contentContainer, 'content')}>
           <p>
-            Earning XP unlocks tiers of veggie badges and eventually earns you new referrals. Each successful referral
-            will contribute Salad Bucks to your balance!
+            Earning XP unlocks tiers of veggie badges.
```
dsarfati: Why is one in a `<p>` and the other not?

PR #157, opened 2019-09-06
Added referral entry component to referral settings DEV-11

dsarfati:


dsarfati:
@thedigitaldesign We should review the way that I used the async calls + `setState`. This is the first time that we truly handle all the different errors that could be returned by the server. It also handles the async API requests in a very compartmentalized way that doesn't leak out into other components. If we like the way this looks, we should start moving new API connections to this sort of pattern and eventually move old API connections over as well.

packages/web-app/src/components/primitives/content/ErrorText.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+// Packages
+import withStyles, { WithStyles } from 'react-jss'
+import classnames from 'classnames'
+
+const styles = (theme: SaladTheme) => ({
+  errorText: {
+    margin: '.25rem',
+    color: theme.red,
+    fontFamily: 'sharpGroteskBook25',
```
thedigitaldesign: 0.3.0 onward there is a font family I added in the SaladTheme, this one is, `theme.fontGroteskBook25`

dsarfati: I just created a new JIRA task to update all fonts since we didn't start with that being included in the theme. DEV-484

PR #166, opened 2019-09-08
DEV-490 & DEV-488

thedigitaldesign:
- Added in an interval that hits loadMachineInfo until we have machineInfo
-- This is a ugly, but is needed to ensure we have set machineInfo.
- Found that the earningRate was tied to this issue. Without machineInfo we couldn't get the user and their earning rate. On refresh we would fetch it, but now it is updated when we get machineInfo from the setInterval
- Added in 'Loading' under Earning Rate while we are fetching machineInfo. Moved the logic off of the .tsx and onto the store

### DIFF
--- a/packages/web-app/src/modules/machine/MachineStore.ts
+++ b/packages/web-app/src/modules/machine/MachineStore.ts
@@ -1,14 +1,29 @@
 import { action, observable, computed } from 'mobx'
 import { Machine } from './models/Machine'
+import { RootStore } from '../../Store'
 
 export class MachineStore {
   @observable
   public currentMachine?: Machine
 
   @computed
-  get currentEarningRate(): number {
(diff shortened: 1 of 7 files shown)


packages/web-app/src/modules/machine/MachineStore.ts
```diff
@@ -1,15 +1,30 @@
...
+import { RootStore } from '../../Store'
 
 export class MachineStore {
   @observable
   public currentMachine?: Machine
 
   @computed
-  get currentEarningRate(): number {
-    return this.currentMachine ? this.currentMachine.earningRate : 0
+  get currentEarningRate(): string {
```
dsarfati: I think this should stay as a number instead of returning a formatted string. We need to use the `currentEarningRate` to compute how long until they earn each reward.

PR #185, opened 2019-09-16
Pointed back to Canny DEV-529

dsorensen00:
Pointed back to Canny


dsarfati:
```
/opt/build/repo/packages/web-app/src/modules/settings-views/SettingsContainer.ts
11:35:16 AM: Type error: 'openLink' is declared but its value is never read.  TS6133
11:35:16 AM:     4 | import { Config } from '../../config'
11:35:16 AM:     5 |
11:35:16 AM:   > 6 | const openLink = (url: string) => {
11:35:16 AM:       |       ^
11:35:16 AM:     7 |   window.open(url, '_blank')
11:35:16 AM:     8 | }
11:35:16 AM:     9 |
```

PR #186, opened 2019-09-17
DEV-520 Mixpanel

thedigitaldesign:
Added Mixpanel tracking into new areas of the app


packages/web-app/src/modules/auth/TokenStore.ts
```diff
@@ -2,7 +2,7 @@ import { action, observable } from 'mobx'
 import { TokenData } from './models'
 import * as Storage from '../../Storage'
 
-export const SALAD_TOKEN = 'NEW_TOKEN'
+const SALAD_TOKEN = 'NEW_TOKEN'
```
dsarfati:
```suggestion
export const SALAD_TOKEN = 'NEW_TOKEN'
```

PR #202, opened 2019-10-07
WIP: DEV-495 Mining Status tooltip

thedigitaldesign:
- When creating a new component for an info tooltip the data of the tooltip was always overwritten by the data in the ALPHA RELEASE tooltip
- Added tooltip code to the UserStatusSummary in an effort to move on from this story


dsarfati:
Waiting on copy update and then we are ready to merge

PR #248, opened 2019-12-16
Additional miners with fallback logic for mining.

TimTinkers:
This update includes two additional miners, Claymore and XMRig. These should be operable for users who are encountering situations where GMiner does not work or work reliably on their machines.

This update includes logic for moving from one miner to another when a failure is encountered. Miners are only rotated in the event that they do not work at all for a user.

We've seen Claymore work in situations where GMiner triggers EPERM issues.


TimTinkers:
Oops, I never should have committed the package lock.

packages/web-app/src/modules/salad-bowl/PluginDefinitionFactory.ts
```diff
@@ -41,6 +44,45 @@ export const beamV2Definition = (machine: Machine): PluginDefinition | undefined
 const beamUser = (location: string, minerId: string) =>
   `-s beamv2.${location}.nicehash.com -n 3378 -u ${miningAddress}.${minerId}`
 
+export const claymoreDefinition = (machine: Machine): PluginDefinition | undefined => {
+  let def = {
+    name: 'Claymore-15',
+    downloadUrl: 'https://github.com/SaladTechnologies/plugin-downloads/releases/download/claymore15/claymore-15-windows.zip',
+    exe: 'EthDcrMiner64.exe',
+    args: `-esm 3 -ewal ${miningAddress}.${machine.minerId} -epool daggerhashimoto.usa.nicehash.com:3353 -allpools 1 -allcoins 0`,
```
dsarfati: We need to add all the nicehash pools as backups. We had issues with the usa stratum before and havent had the issues since we added fallback pools

packages/web-app/src/modules/salad-bowl/PluginDefinitionFactory.ts
```diff
@@ -41,6 +44,45 @@ export const beamV2Definition = (machine: Machine): PluginDefinition | undefined
...
+
+  return def
+}
+
+export const xmrigDefinitionCuda = (machine: Machine): PluginDefinition | undefined => {
+  let def = {
+    name: 'XMRig-5.2.0-CUDA',
+    downloadUrl: 'https://github.com/SaladTechnologies/plugin-downloads/releases/download/xmrig-5.2.0/xmrig-5.2.0-windows-cuda.zip',
+    exe: 'xmrig.exe',
+    args: `--donate-level 1 --no-cpu --opencl --cuda -o stratum+tcp://randomxmonero.usa.nicehash.com:3380 -u ${miningAddress} -k --nicehash --coin monero`,
```
dsarfati: Add backup pools

packages/web-app/src/modules/salad-bowl/PluginDefinitionFactory.ts
```diff
@@ -41,6 +44,45 @@ export const beamV2Definition = (machine: Machine): PluginDefinition | undefined
...
+
+  return def
+}
+
+export const xmrigDefinitionOpenCL = (machine: Machine): PluginDefinition | undefined => {
+  let def = {
+    name: 'XMRig-5.2.0-OpenCL',
+    downloadUrl: 'https://github.com/SaladTechnologies/plugin-downloads/releases/download/xmrig-5.2.0/xmrig-5.2.0-windows-opencl.zip',
+    exe: 'xmrig.exe',
+    args: `--donate-level 1 --no-cpu --opencl --cuda -o stratum+tcp://randomxmonero.usa.nicehash.com:3380 -u ${miningAddress} -k --nicehash --coin monero`,
```
dsarfati: Add backup pools

packages/web-app/src/modules/salad-bowl/SaladBowlStore.ts
```diff
@@ -30,12 +30,20 @@ export class SaladBowlStore {
 
   @computed
   get isRunning(): boolean {
-    return this.plugin.status !== PluginStatus.Stopped && this.plugin.status !== PluginStatus.Unknown
+    if (this.plugins[0]) {
+      return this.plugins[0].status !== PluginStatus.Stopped && this.plugins[0].status !== PluginStatus.Unknown
+    } else {
+      return false
+    }
```
dsarfati: If it is always using index 0, won't the status be incorrect if it is the 2nd or 3rd miner that ends up running?

packages/web-app/src/modules/salad-bowl/SaladBowlStore.ts
```diff
@@ -114,17 +131,28 @@ export class SaladBowlStore {
...
+      }
+      plugins.push({
+        name: pluginName,
+        status: PluginStatus.Initializing
+      })
+    }
+    this.plugins = plugins
 
-    yield this.store.native.send('start-salad', pluginDefinition)
+    yield this.store.native.send('start-salad', pluginDefinitions)
```
dsarfati: We will need to check the bridge api version and only send 1 plugin definition to older desktop apps that don't support the array of plugins

packages/web-app/src/modules/salad-bowl/models/ErrorCategory.ts
```diff
@@ -4,4 +4,5 @@ export enum ErrorCategory {
   AntiVirus = 'antiVirus',
   Network = 'network',
   Silent = 'silent',
+  Incompatible = 'incompatible'
```
dsarfati: Let's change to to be a more unique error name since incompatible is something we use in other places related to not being on the whitelist. Maybe PluginManager or something related to plugins??

packages/web-app/src/modules/salad-bowl/SaladBowlStore.ts
```diff
@@ -83,6 +97,9 @@ export class SaladBowlStore {
       case ErrorCategory.AntiVirus:
         this.store.ui.showModal('/errors/anti-virus')
         break
+      case ErrorCategory.Incompatible:
+        this.store.ui.showModal('/errors/no-plugins')
```
dsarfati: We want to check for the category and the error code here since this modal will be specifically for when we run out of plugins to try.

seniorquico:
Superseded by #256.

PR #251, opened 2019-12-18
In app automated banner DEV-849

dsorensen00:


packages/web-app/src/modules/home-views/components/NotificationBanner.tsx
```diff
@@ -17,20 +18,24 @@ const styles = (theme: SaladTheme) => ({
...
-  endDate?: Date
-  text?: string
+  bannerInfo?: BannerInfo
 }
 
 class _NotificationBanner extends Component<Props> {
   render() {
-    const { now, startDate, endDate, text, classes } = this.props
+    const { bannerInfo, classes } = this.props
+    let now = new Date()
```
dsarfati: This `now` will only happen when a render occurs, this will not automatically update to the current time. I would expect render to occur when the app opens and then would likely not happen again. If you wanted to keep checking the times, then you will need to start a timer in the `componentWillMount` and kill that timer in the `componentdDidUnmount` that will set the state of now every X sec, to force this page to re render. I just something similar for the dynamic earning history to get the labels to update every 1 minute, if you would like an example.

packages/web-app/src/modules/home/HomeStore.ts
```diff
@@ -0,0 +1,21 @@
...
+  @action.bound
+  loadBannerInfo = flow(function*(this: HomeStore) {
+    try {
+      let res = yield this.axios.get<BannerInfo>('/notification-banner')
+      this.bannerInfo = res.data
+    } catch (err) {
+      console.log(err)
+    }
+  })
+//@ts-ignore
```
dsarfati: What is being ignored??

packages/web-app/src/modules/home/models/BannerInfo.ts
```diff
@@ -0,0 +1,6 @@
+export interface BannerInfo
+{
+  startDate: string
+  endDate:string
```
dsarfati: Can these be more strongly typed using either `Date` or something in moment??

packages/web-app/yarn.lock
```diff
@@ -1091,29 +1091,29 @@
...
+  integrity sha512-CcM/fIFwZlRdiWG/25xE/wHbtyUuCtqoCTrr6BsWw7hH072fR++n4L56KPydAr3ANgMJMjT8v83ZFIsDc7kE+A==
 
-"@fortawesome/fontawesome-svg-core@^1.2.14":
-  version "1.2.25"
-  resolved "https://registry.yarnpkg.com/@fortawesome/fontawesome-svg-core/-/fontawesome-svg-core-1.2.25.tgz#24b03391d14f0c6171e8cad7057c687b74049790"
-  integrity sha512-MotKnn53JKqbkLQiwcZSBJVYtTgIKFbh7B8+kd05TSnfKYPFmjKKI59o2fpz5t0Hzl35vVGU6+N4twoOpZUrqA==
+"@fortawesome/fontawesome-svg-core@^1.2.25":
+  version "1.2.26"
+  resolved "https://registry.yarnpkg.com/@fortawesome/fontawesome-svg-core/-/fontawesome-svg-core-1.2.26.tgz#671569271d6b532cdea5e3deb8ff16f8b7ac251d"
+  integrity sha512-3Dfd/v2IztP1TxKOxZiB5+4kaOZK9mNy0KU1vVK7nFlPWz3gzxrCWB+AloQhQUoJ8HhGqbzjliK89Vl7PExGbw==
```
dsarfati: Did you update these? I don't see any updates to the package.json file.

packages/web-app/src/Store.ts
```diff
@@ -111,4 +114,8 @@ export class RootStore {
     this.analytics.trackLogout()
     this.saladBowl.stop()
   }
+
+  loadData = () =>{
+    this.home.loadBannerInfo()
+  }
```
dsarfati: I think this should be in the refresh service instead. That way it will pull the most up to date info.

packages/web-app/src/modules/home/HomeStore.ts
```diff
@@ -0,0 +1,21 @@
...
+  @observable
+  public bannerInfo?: BannerInfo
+
+  @action.bound
+  loadBannerInfo = flow(function*(this: HomeStore) {
+    try {
+      let res = yield this.axios.get<BannerInfo>('/notification-banner')
+      this.bannerInfo = res.data
+    } catch (err) {
+      console.log(err)
```
dsarfati: We should probably set `this.bannerInfo = undefined` if the request fails. That way we always clear old banner data.

packages/web-app/yarn.lock
dsorensen00: @dsarfati these changes were applied when I ran the yarn command to install everything, I assumed these were updates?

packages/web-app/src/Store.ts
dsorensen00: I totally forgot to delete that XD good catch

packages/web-app/src/modules/home/models/BannerInfo.ts
dsorensen00: I can look into it, but I don't believe there is a way of better typing it especially with moment.js. The Date type may be able to store the C# DateTimeOffset?

packages/web-app/src/modules/home/HomeStore.ts
dsorensen00: Because `store` was unused and was causing a runtime error not allowing it to run.

packages/web-app/src/modules/home-views/components/NotificationBanner.tsx
dsorensen00: Yeah I was kinda struggling with this one. You're suggesting that I have a timer that somehow triggers another render every X seconds? Would love to know where your example is to take a look.

dsorensen00: @dsarfati I just tested this and it does refresh within the minute without a forced re-render, going to mark this as resolved.

packages/web-app/yarn.lock
seniorquico: I think there's a config problem here. I'll investigate and try to provide some notes. I had to update the lock files in my other PR (even without any changes to package.json) to resolve Netlify build problems. May want to pull in the latest bits from master.

packages/web-app/src/modules/home/HomeStore.ts
dsarfati: So shouldn't you just remove store than?

PR #253, opened 2019-12-19
Enabling XMRig as the active mining plugin.

TimTinkers:
Taking a shortcut from the full fallback PR to try and fix the present NiceHash issues.


seniorquico:
Superseded by #258.

PR #258, opened 2019-12-20
Add XMRig as a fallback miner DEV-915

seniorquico:
This update also includes an additional miner, XMRig. This should be operable for users who are encountering situations where both GMiner and Claymore does not work or work reliably on their machines.

### DIFF
--- a/packages/web-app/src/modules/salad-bowl/definitions/getClaymoreDefinition.ts
+++ b/packages/web-app/src/modules/salad-bowl/definitions/getClaymoreDefinition.ts
@@ -2,14 +2,20 @@ import { Machine } from '../../machine/models/Machine'
 import { PluginDefinition } from '../models'
 import { MINING_ADDRESS, STANDARD_ERRORS } from './constants'
 
+const claymoreRegion = (location: string) => `-epool daggerhashimoto.${location}.nicehash.com:3353`
+
 export const getClaymoreDefinition = (machine: Machine): PluginDefinition => {
   let def = {
     name: 'Claymore-15',
     downloadUrl:
       'https://github.com/SaladTechnologies/plugin-downloads/releases/download/claymore15/claymore-15-windows.zip',
     exe: 'EthDcrMiner64.exe',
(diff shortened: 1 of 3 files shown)


packages/web-app/src/modules/salad-bowl/definitions/getClaymoreDefinition.ts
```diff
@@ -2,14 +2,20 @@ import { Machine } from '../../machine/models/Machine'
...
       'https://github.com/SaladTechnologies/plugin-downloads/releases/download/claymore15/claymore-15-windows.zip',
     exe: 'EthDcrMiner64.exe',
-    args: `-esm 3 -ewal ${MINING_ADDRESS}.${machine.minerId} -epool daggerhashimoto.usa.nicehash.com:3353 -allpools 1 -allcoins 0`,
-    runningCheck: 'ETH: GPU0 [1-9]+(.[0-9][0-9][0-9]?)? [KMG]h/s',
+    args: `-esm 3 -ewal ${MINING_ADDRESS}.${machine.minerId} ${claymoreRegion('usa')} ${claymoreRegion(
+      'eu',
+    )} ${claymoreRegion('hk')} ${claymoreRegion('jp')} ${claymoreRegion('in')} ${claymoreRegion(
+      'br',
+    )} -allpools 1 -allcoins 0`,
+    runningCheck: 'ETH: GPU0 [1-9]+(\\.[0-9][0-9][0-9]?)? [KMG]h/s',
```
dsarfati: Having this include the `GPU0` might cause issues if we have a machine with multiple GPUs (aka `GPU1` is running instead).

seniorquico: Should the RegExp be relaxed?

```suggestion
    runningCheck: 'ETH: GPU[0-9]+ [1-9]+(\\.[0-9][0-9][0-9]?)? [KMG]h/s',
```

seniorquico:
Superseded by #260.

Note: The Claymore changes (fallback to multiple NiceHash pools and a fix to the `runningCheck`) will be reintroduced on an upcoming PR.

PR #261, opened 2019-12-21
Add Claymore fallback pools and improve fast failure scenarios DEV-915

seniorquico:
This adds fallback pools when using Claymore. This also reduces the time it takes for pool failures to be detected when using Gminer.

### DIFF
--- a/packages/desktop-app/src/salad-bowl/Plugin.ts
+++ b/packages/desktop-app/src/salad-bowl/Plugin.ts
@@ -236,9 +236,13 @@ export class Plugin {
 
     if (this.pluginDefinition.autoRestart === false) {
       console.log(`Auto restart is disabled for ${this.name}`)
+      if (!stopped) {
+        await this.stop()
+      }
+
       return
     }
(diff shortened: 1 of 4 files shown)


seniorquico:
I found a bug in the desktop app code. Namely, the `stopCalled` flag gets set to `true` when a plugin is auto-restarted. This has the unfortunate effect of only restarting once and then stopping. Eventually, the initial timeout watchdog will fire (after 10 minutes) and fallback to the next miner. However, if a miner ever successfully starts, the watchdog timer is cancelled. In this scenario, if the miner ever fails later, it will only survive a single auto-restart. After the first restart, the user will be stuck in a state without a miner running.

The only fix is to set `stopCalled` to `false` in the `start` method. However, this requires a desktop app update--a non-trivial operation for end users with the new code-signing certificate. I do not see a workaround with a web app change. If we set every plugin definition's `autoRestart: false`, we will get _no_ miner status change events. (This is yet _another_ bug.)

PR #264, opened 2020-01-04
Add Nanopool backup plugin definition DEV-944

seniorquico:
This re-introduces a plugin definition that points to Nanopool. This exists as a fallback option in the event NiceHash has complete pool outages.


packages/web-app/src/modules/salad-bowl/definitions/constants.ts
```diff
@@ -1,6 +1,7 @@
 import { ErrorAction } from '../models/ErrorAction'
 import { ErrorCategory } from '../models/ErrorCategory'
 
+export const WALLET_ADDRESS = '0x6ff85749ffac2d3a36efa2bc916305433fa93731'
```
dsarfati: Which wallet is this? Nanopool or our kraken wallet?

seniorquico:
I'm going to keep this in the WIP holding pattern. It's ready to go, but we really don't need it in the absence of NiceHash, Ethermine, and Flypool issues. The backend is undergoing a significant overhaul, and this may be declined once those changes are live.

PR #333, opened 2020-03-15
Filter plugin definitions for lower-end GPUs

seniorquico:
Additional GPUs have been added to Salad, and some of them have limited VRAM. For these GPUs, unsupported plugin definitions should be filtered from the fallback list.


seniorquico:
@dsarfati and @thedigitaldesign, I'm unable to test the Netlify deploy preview locally by updating the desktop app config. Signing in with Auth0 fails on the Salad login API with a 400 Bad Request. The Auth0 login page is titled "Salad Test - Old".

dsarfati:
@seniorquico I will update all the env files now that the test env is back up and running. I had the same issue when trying to run locally as well so I pointed everything to the prod env.

dsarfati:
@seniorquico Just tested running the web-app locally and it looks like the `.env.development` file is working as expected. I was able to log into the test env. It looks like the `REACT_APP_AUTH0_CLIENT_ID` in the `netlify.toml` is pointed to the wrong auth0 tenant for the preview deployments. Updating that in a PR right now.

PR #346, opened 2020-03-27
Typo fix for referral panel

VukAnd:
*disclaimer: First time contributing here. I have absolutely no idea if I should be the one doing this or if it actually works, because the instructions in the README file don't seem to be working (and the CONTRIBUTING file is nonexistent) - so sorry in advance if I messed up something here


dsarfati:
Thanks for the PR @VukAnd! You are right that we don't do a good job of a contributing guide.

PR #347, opened 2020-03-27
Added default retry logic to all API requests

dsarfati:


dsarfati:
@seniorquico I found a very very popular addon for axios that does it for us. What do you think about using this for now and seeing what happens? This will apply to all requests. I know you spent a lot of time looking into retry logic and patterns so didn't know if you had better recommendations.

packages/web-app/src/axiosFactory.ts
```diff
@@ -21,6 +22,8 @@ export const createClient = (): AxiosInstance => {
     },
   )
 
+  axiosRetry(c, { retries: 5, retryDelay: exponentialDelay })
```
seniorquico: Please review and consider the following:

```ts
import axios, { AxiosError, AxiosInstance } from 'axios'
import axiosRetry, { exponentialDelay } from 'axios-retry/lib'
import isRetryAllowed from 'is-retry-allowed'

/**
 * The list of safe HTTP request methods. HTTP requests using these methods may be retried.
 */
const SAFE_HTTP_METHODS: ReadonlyArray<string> = ['get', 'head', 'options']

/**
 * Determines whether a HTTP request may be retried. HTTP requests may be retried on network errors. HTTP requests using
 * safe methods (GET, HEAD, and OPTIONS) may be retried after a HTTP response with status codes 408 or 500-599.
 * @param error The error.
 * @returns `true` if the HTTP request may be retried; otherwise, `false`.
 */
const shouldRetryDownload = (error: any): boolean => {
  if (error.isAxiosError === true) {
    const axiosError: AxiosError<any> = error
    if (
      SAFE_HTTP_METHODS.indexOf(axiosError.config.method.toLowerCase()) !== -1 &&
      (axiosError.response == null ||
        axiosError.response.status === 408 ||
        (axiosError.response.status >= 500 && axiosError.response.status <= 599))
    ) {
      return true
    }
  }

  return isRetryAllowed(error)
}

/**
 * Creates a new HTTP client. The client will retry all HTTP requests on network errors and HTTP requests using safe
 * methods (GET, HEAD, and OPTIONS) after a HTTP response with status codes 408 or 500-599. The client will retry up to
 * three times with a default timeout of 10 seconds.
 * @returns The HTTP client.
 */
export const createHttpClient = (): AxiosInstance => {
  const httpClient = axios.create()
  httpClient.defaults.timeout = 10000
  axiosRetry(httpClient, {
    retryDelay: exponentialDelay,
    retryCondition: shouldRetryDownload,
    shouldResetTimeout: true,
  })
  return httpClient
}
```

dsarfati: Done

PR #349, opened 2020-03-27
Move the offerwall fine print

VukAnd:
Considering a part of the "fine print" is mentioned at the top with the offerwall enabled,
image
it's probably a good idea to move the fine print higher up in the "disabled" view (consistency!)


VukAnd:
Closing this after further consideration - if the change looks good you can open and merge however

PR #379, opened 2020-04-20
Added PhoenixMiner as a new Ethereum mining option

TimTinkers:
PhoenixMiner is also preferred over Claymore when mining Ethereum on compatible cards.

### DIFF
--- a/packages/web-app/src/modules/salad-bowl/PluginDefinitionFactory.ts
+++ b/packages/web-app/src/modules/salad-bowl/PluginDefinitionFactory.ts
@@ -4,6 +4,9 @@ import { getCCMinerX16RDefinition } from './definitions/getCCMinerX16RDefinition
 import { getClaymoreEthashBitflyDefinition } from './definitions/getClaymoreEthashBitflyDefinition'
 import { getClaymoreEthashDefinition } from './definitions/getClaymoreEthashDefinition'
 import { getClaymoreEthashNanopoolDefinition } from './definitions/getClaymoreEthashNanopoolDefinition'
+import { getPhoenixMinerEthashBitflyDefinition } from './definitions/getPhoenixMinerEthashBitflyDefinition'
+import { getPhoenixMinerEthashDefinition } from './definitions/getPhoenixMinerEthashDefinition'
+import { getPhoenixMinerEthashNanopoolDefinition } from './definitions/getPhoenixMinerEthashNanopoolDefinition'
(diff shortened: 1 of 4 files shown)


packages/web-app/src/modules/salad-bowl/PluginDefinitionFactory.ts
```diff
@@ -27,10 +30,14 @@ export const getPluginDefinitions = (store: RootStore): PluginDefinition[] => {
 
   // Ethereum / Ethash
   if (preferNiceHash && has4gbSupport) {
+    pluginDefinitions.push(getPhoenixMinerEthashDefinition(machine)) // NiceHash
     pluginDefinitions.push(getClaymoreEthashDefinition(machine)) // NiceHash
```
seniorquico, line 34: Why should we keep Claymore in the list?

packages/web-app/src/modules/salad-bowl/definitions/getPhoenixMinerEthashNanopoolDefinition.ts
```diff
@@ -0,0 +1,19 @@
+import { Machine } from '../../machine/models/Machine'
+import { PluginDefinition } from '../models'
+import { STANDARD_ERRORS, ETH_WALLET_ADDRESS } from './constants'
+
+export const getPhoenixMinerEthashNanopoolDefinition = (machine: Machine): PluginDefinition => {
+  let def = {
+    name: 'PhoenixMiner-Nanopool-Ethash',
```
seniorquico, line 7: All of these plugins have different names, which will currently lead to multiple downloads/copies installed. This is problematic for users' AV/firewalls. Can we either prioritize that rework, or can we make these (temporarily) use the same name?

packages/web-app/src/modules/salad-bowl/PluginDefinitionFactory.ts
TimTinkers: It's still works quite well for most of our users and would make a good option to fall back to.

packages/web-app/src/modules/salad-bowl/definitions/getPhoenixMinerEthashNanopoolDefinition.ts
TimTinkers: All are renamed to PhoenixMiner now.

PR #407, opened 2020-05-06
Added new veggies, re-balanced XP

dsarfati:
https://github.com/SaladTechnologies/salad-platform/issues/546


dsarfati:
@VukAnd Now with 38% more pantry

PR #498, opened 2020-08-23
Nice small changes

Vukky123:
5b22875 - 👥 RewardHeaderBar now displays Donate Now if the reward is a donation.
* ~~I attempted to change Order Summary to Donation Summary and Pay With Salad to Donate With Salad for donations too, however, I couldn't figure it out. Does a dev know how to do this?~~ [see this comment](https://github.com/SaladTechnologies/salad-applications/pull/498#issuecomment-680052473)

87d0560 - Added Origin 👥, Xbox, and Apple as platforms to RewardInfoPanel and RewardPlatform.

Changes marked with :busts_in_silhouette: are @stabbed's idea!
*This is my first code contribution. Do not be surprised if this PR is horrible.*


packages/web-app/src/modules/reward-views/components/RewardHeaderBar.tsx
```diff
@@ -111,6 +111,7 @@ class _RewardHeaderBar extends Component<Props> {
   render() {
     const { reward, classes, ...rest } = this.props
 
+    let donation = reward?.tags.includes('donations')
```
seniorquico, line 114: The tags are not always the same case. Instead we should match against the `category` which will remain consistent:

```suggestion
    let donation = reward?.category === "donation"
```

seniorquico:
Thanks, @Vukky123. The Salad Pay labels will have to remain as-is for technical and branding purposes. Otherwise, it looks good to me aside from the one requested change. If you get this updated, I can run it through QA testing.

seniorquico:
@Vukky123 Is this ready for QA testing?

Vukky123:
Should be ready now.

PR #526, opened 2020-09-16
Move icon path definitions out from main.ts into icons.ts.

Krish120003:
As of https://trello.com/c/bz057Sg7/1-icon-compatibility

`icons.ts` uses the `process` module to check for platform. `.ico` files are used on win32 and `.png` for linux/darwin.

### DIFF
--- a/packages/desktop-app/src/icons.ts
+++ b/packages/desktop-app/src/icons.ts
@@ -5,8 +5,6 @@ declare const __static: string
 
 export const LOGO_PATH = path.join(__static, process.platform === 'win32' ? 'logo.ico' : 'logo.png')
 
-export const LOGO_ANIMATED_PATH = path.join(__static, 'logo.gif')
-
 export const TRAY_ICON_PATH = LOGO_PATH
 
 export const TRAY_ACTIVE_ICON_PATH = path.join(
--- a/packages/desktop-app/src/icons.ts
+++ b/packages/desktop-app/src/icons.ts
@@ -19,6 +19,11 @@ export const TRAY_ALERT_ICON_PATH = path.join(
   process.platform === 'win32' ? 'logo-alert.ico' : 'logo-alert.png',
 )
 
+// For MacOS Dock
+export const DOCK_LOGO_ACTIVE_PATH = path.join(__static, 'logo-active.png')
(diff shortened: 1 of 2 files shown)


packages/desktop-app/src/icons.ts
```diff
@@ -0,0 +1,29 @@
...
+  __static,
+  process.platform === 'win32' ? 'logo-alert.ico' : 'logo-alert.png',
+)
+
+// [Placeholder] For MacOS Dock
+export const DOCK_LOGO_ACTIVE_PATH = path.join(__static, 'logo-active.png')
+// [Placeholder] For MacOS Dock
+export const DOCK_LOGO_ALERT_PATH = path.join(__static, 'logo-alert.png')
+
+export const TASKBAR_OVERLAY_ACTIVE_PATH = path.join(__static, 'taskbar-overlay-active.png')
```
seniorquico, line 27:
```suggestion
export const TAKSBAR_ACTIVE_OVERLAY_ICON_PATH = path.join(__static, 'taskbar-overlay-active.png')
```

packages/desktop-app/src/icons.ts
```diff
@@ -0,0 +1,29 @@
+import path from 'path'
+import process from 'process'
+
+declare const __static: string
+
+export const LOGO_PATH = path.join(__static, process.platform === 'win32' ? 'logo.ico' : 'logo.png')
+
+export const LOGO_ANIMATED_PATH = path.join(__static, 'logo.gif')
```
seniorquico, line 8: Let's leave out any assets that aren't (yet) used in the code. They can be added later, when they're needed. (We frequently check for "unused exports", and these will all be flagged by the tooling as unnecessary.)

seniorquico (changes requested):
Thanks, @Krish120003! I have just a few comments, but this is a great refactor.

packages/desktop-app/src/icons.ts
Krish120003: The LOGO_PATH is being used for the offlineWindow's icon. I have removed the LOGO_ANIMATED_PATH since it is not being used.

seniorquico (approved):

PR #530, opened 2020-09-17
Notification support for linux/darwin

Krish120003:
Add notification support for linux/darwin using electron's Notification. Untested on darwin, but should work out of the box.

### DIFF
--- a/packages/desktop-app/src/main.ts
+++ b/packages/desktop-app/src/main.ts
@@ -1,5 +1,16 @@
 import * as Sentry from '@sentry/electron'
-import { app, BrowserWindow, Input, ipcMain, Menu, nativeImage, powerMonitor, shell, Tray } from 'electron'
+import {
+  app,
+  BrowserWindow,
+  Input,
+  ipcMain,
+  Menu,
+  nativeImage,
+  Notification,
+  powerMonitor,
+  shell,
+  Tray,
+} from 'electron'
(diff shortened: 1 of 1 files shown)


packages/desktop-app/src/main.ts
```diff
@@ -366,19 +377,30 @@ const createMainWindow = () => {
...
+        },
+        (err) => {
+          if (err) {
+            console.warn('Notification error')
+            console.warn(err)
+          }
+        },
+      )
+    } else {
+      console.log('creating new notification', message)
```
seniorquico, line 395: Remove debug code.

packages/desktop-app/src/main.ts
```diff
@@ -366,19 +377,30 @@ const createMainWindow = () => {
...
+          appID: 'salad-technologies-desktop-app',
+        },
+        (err) => {
+          if (err) {
+            console.warn('Notification error')
+            console.warn(err)
+          }
+        },
+      )
+    } else {
```
seniorquico, line 394: Let's check to ensure notifications are supported:

```suggestion
    } else if (Notification.isSupported()) {
```

Docs: https://www.electronjs.org/docs/api/notification#notificationissupported

seniorquico (changes requested):

seniorquico (approved):

PR #531, opened 2020-09-18
Add build config for linux, add build icons

Krish120003:
Still need to add the 512x512 icon into the build/icons folder.


seniorquico:
@Krish120003 The Linux package assets have been merged into the `master` branch.

Krish120003 (commented):
Add build config for linux

packages/desktop-app/electron-builder.json
```diff
@@ -16,5 +16,12 @@
     "artifactName": "${productName} Installer ${version}.${ext}",
     "license": "LICENSE.md",
     "oneClick": false
+  },
+  "linux": {
+    "target": ["deb", "rpm", "snap"],
+    "category": "Utility"
+  },
+  "rpm": {
+    "depends": ["libappindicator1", "libnotify-bin"]
```
seniorquico, line 25: Do we know that explicitly declaring the dependencies is required? Do they not automatically include these dependencies in their template?

packages/desktop-app/electron-builder.json
```diff
@@ -16,5 +16,12 @@
     "artifactName": "${productName} Installer ${version}.${ext}",
     "license": "LICENSE.md",
     "oneClick": false
+  },
+  "linux": {
+    "target": ["deb", "rpm", "snap"],
```
seniorquico, line 21: Let's also include the AppImage target:

```suggestion
    "target": ["AppImage", "deb", "rpm", "snap"],
```

I'm not sure how popular this format is, but it was easy to download & run to app.

seniorquico (changes requested):

packages/desktop-app/electron-builder.json
seniorquico: I created a `.rpm` build without explicitly declaring dependencies, and I tested it on CentOS 8. I was able to confirm that it does _not_ include `libappindicator1` as a dependency. The necessary dependency appears to be `libappindicator-gtk3` (e.g. Chrome & Slack depend on this package).

PR #557, opened 2020-10-10
Add clinfo along with default dependencies for linux build

Krish120003:
Does not add `clinfo` as a dependency with the `.snap` package


seniorquico:
Declining this in favor of the changes going into the upstream [`systeminformation` package](https://github.com/sebhildebrandt/systeminformation/pull/419). Thanks for all your hard work on this, @Krish120003!

PR #561, opened 2020-10-19
Change Promo to Referral

Fvn556:
If I did a dumb then that's cause I don't know what I'm doing when it comes to github


Walker21390 (approved):
Hey Fvn, I'm here to tell you that you didn't do a dumb!

dsarfati:

PR #574, opened 2020-11-18
Fixed the mobile offerwall bug

Fvn556:
you know the one where you couldnt really do anything other than stay on the offerwalls page on mobile


dsarfati:

PR #575, opened 2020-11-18
Add Zendesk authentication

scottcinatl:
Adds authentication callback to Zendesk web widget settings by passing along auth user JWT. This also moves the initialization of the Zandesk snippet from `index.html` to `Store.ts`.

### DIFF
--- a/packages/web-app/src/Store.ts
+++ b/packages/web-app/src/Store.ts
@@ -61,7 +61,7 @@ export class RootStore {
   public readonly vault: VaultStore
   public readonly version: VersionStore
   public readonly engagement: EngagementStore
-  private readonly zendesk : Zendesk;
+  private readonly zendesk: Zendesk;
 
   constructor(readonly axios: AxiosInstance) {
     this.routing = new RouterStore()
@@ -90,10 +90,9 @@ export class RootStore {
     // Start refreshing data
     this.refresh.start()
 
-    // initialize Zendesk
-    this.zendesk.intializeZendesk();
-
     autorun(() => {
+      // initialize Zendesk
+      this.zendesk.intializeZendesk();
       if (this.auth.isAuthenticated) {
         this.onLogin()
       } else {
--- a/packages/web-app/src/Store.ts
+++ b/packages/web-app/src/Store.ts
@@ -20,8 +20,8 @@ import { StopReason } from './modules/salad-bowl/models'
 import { VaultStore } from './modules/vault'
 import { VersionStore } from './modules/versions'
 import { ExperienceStore } from './modules/xp'
-import { UIStore } from './UIStore'
 import { Zendesk } from './modules/zendesk'
+import { UIStore } from './UIStore'
 
 configure({
   // computedRequiresReaction: process.env.NODE_ENV === 'development',
@@ -61,7 +61,7 @@ export class RootStore {
   public readonly vault: VaultStore
   public readonly version: VersionStore
   public readonly engagement: EngagementStore
-  private readonly zendesk: Zendesk;
+  private readonly zendesk: Zendesk
 
   constructor(readonly axios: AxiosInstance) {
     this.routing = new RouterStore()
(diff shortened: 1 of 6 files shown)


scottcinatl:
@seniorquico Verified that the token I was receiving was for the correct account, but I noticed Zendesks own calls to verify the authentication was denied due to the request coming from a unknown domain (local).

Capture

So I am not sure how to go about verification on the Zendesk side of things that requests are being properly associated with the user account.

packages/web-app/src/Store.ts
```diff
@@ -81,13 +83,16 @@ export class RootStore {
...
     this.engagement = new EngagementStore(this)
+    this.zendesk = new Zendesk(this, axios)
 
     addAuthInterceptor(axios, this.auth)
 
     // Start refreshing data
     this.refresh.start()
 
     autorun(() => {
+      // initialize Zendesk
```
seniorquico, line 94: nit: The comment looks redundant; the function name describes what's going on.

packages/web-app/src/Store.ts
```diff
@@ -104,22 +109,8 @@ export class RootStore {
...
-            value: profile.email.toLocaleLowerCase(),
-            readOnly: true, // optional
-          },
-        })
-      } catch (e) {
-        console.error('Unable to prefill Zendesk')
-        console.error(e)
-      }
+      // prefill profile data to zendesk
+      this.zendesk.prefillProfile(profile.username, profile.email)
```
seniorquico, line 113: nit: The comment looks redundant; the field & function names describe what's going on.

packages/web-app/src/Store.ts
```diff
@@ -81,13 +83,16 @@ export class RootStore {
...
+    this.zendesk = new Zendesk(this, axios)
 
     addAuthInterceptor(axios, this.auth)
 
     // Start refreshing data
     this.refresh.start()
 
     autorun(() => {
+      // initialize Zendesk
+      this.zendesk.intializeZendesk();
```
seniorquico, line 95: The only observable used within this `autorun` block is the `isAuthenticated` flag. This changes when the user logs in & logs out. It's possible `initializeZendesk` will be called multiple times in a single app session. As it stands now, this will result in creating multiple `<script ...>` tags to load the Zendesk web widget (one at app startup & one every time the user logs in or logs out). This should probably be split... initializing `window.zESettings` and injecting the `<script ...>` tag should only happen once for the entire app session. This `autorun` block can then be used to react to changes in the authentication state.

packages/web-app/src/modules/zendesk/Zendesk.ts
```diff
@@ -0,0 +1,93 @@
...
+      } catch (e) {
+        console.log("Unable to reauthenticate Zendesk web widget");
+        console.log(e)
+      }
+    }
+  }
+
+  prefillProfile(username: string, email: string) {
+    try {
+      window.zE && window.zE('webWidget', 'prefill', {
```
seniorquico, line 66: What happens if the Zendesk web widget scripts haven't loaded by the time this function is called? Wouldn't this just be skipped and the form left with empty values/defaults?

Maybe a better question... is this still necessary when we're using the authentication? It may very well be, but it kinda feels redundant given we've already passed along a signed JWT that contained this info.

seniorquico (changes requested):

packages/web-app/src/modules/zendesk/Zendesk.ts
scottcinatl: With my latest [commit](https://github.com/SaladTechnologies/salad-applications/pull/575/commits/a432d4925a8bcaa874172d5d6ca5e5f0e6a5f953), I moved the inject of the Zendesk to always fire before any authentication process occurs, which includes the prefilling of the form. So there should never be a case when a prefill is attempted by an unauthenticated user. 

After looking around in Zendesk's docs, it does not seem that submitting a successful token triggers and auto filling of form fields. So the built in prefill function seems to be the way to go. As for that information, are you suggesting that I should be getting that info by decoding the JWT instead of getting the information from the store?

packages/web-app/src/modules/zendesk/Zendesk.ts
```diff
@@ -0,0 +1,150 @@
...
+      if (!Zendesk.injectionError) {
+        this.initializeRetryTimeout = setTimeout(() => {
+          this.initializeRetryTimeout = undefined
+          this.logout();
+        }, 1000);
+      }
+
+      return
+    }
+
```
seniorquico, line 138: This should clear any pending "reauthenticate" timer before calling the logout function (we know it won't succeed, so just bail).


```ts    if (this.reauthenticateRetryTimeout) {
      clearTimeout(this.reauthenticateRetryTimeout)
      this.reauthenticateRetryTimeout = undefined
    }```

seniorquico (changes requested):

packages/web-app/src/modules/zendesk/Zendesk.ts
seniorquico: Retry logic added in 9532ea4.

seniorquico (approved):

PR #576, opened 2020-11-23
Offer Toro -> OfferToro (MobileOfferwallPage.tsx)

Vukky123:


dsarfati:

PR #577, opened 2020-11-23
Change OfferToro Link to offertoro from offer-toro

Fvn556:
this was requested by vukky to me btw


dsarfati:

PR #586, opened 2020-12-11
Fixed overlap of tick labels and updated 30 Day x-axis label on Earning Chart

scottcinatl:
- Fixes overlapping x-axis label on Earning Chart
- Updated 30 Day x-axis label to display date instead of time

screenshot with updated color

mobile-ticks

### DIFF
--- a/packages/web-app/src/modules/earn-views/components/EarningChart.tsx
+++ b/packages/web-app/src/modules/earn-views/components/EarningChart.tsx
@@ -1,15 +1,15 @@
-import moment from 'moment'
+import classnames from 'classnames'
 import { uniq } from 'lodash'
+import moment from 'moment'
 import React, { Component } from 'react'
 import withStyles, { WithStyles } from 'react-jss'
-import classnames from 'classnames'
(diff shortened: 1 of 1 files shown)


packages/web-app/src/modules/earn-views/components/EarningChart.tsx
```diff
@@ -45,6 +45,20 @@ const styles = (theme: SaladTheme) => ({
...
+    display: 'block',
+  },
+  '@media screen and (min-width: 1025px)': {
+    tickDesktop: {
+      display: 'block',
+    },
+    tickMobile: {
+      display: 'none',
+    },
+  },
```
seniorquico, line 61: This looks like a nice, quick approach. But how well does it scale across different screen sizes and, in particular, different DPI settings? Were you able to test in multiple configurations to ensure they don't get scrunched together? Is there no other responsive approach, like a continuous scaling up & down as the sizes/DPIs change?

scottcinatl: I just walked through the different DPI settings that we discussed earlier today. The current media query holds up at the different settings and maintains scrunching of the tick lables from occurring.

seniorquico (approved):

PR #590, opened 2020-12-17
Desktop version check and LogOut button fixes

scottcinatl:
1. Only shows LogOut button on Settings page if user is Authenticated
2. Only check desktop version if on the native app


packages/web-app/src/modules/settings-views/SettingsContainer.ts
```diff
@@ -49,7 +49,8 @@ const mapStoreToProps = (store: RootStore): any => {
     appBuild: config.appBuild,
     latestDesktop: store.version.onLatestDesktop,
     onDownloadLatestDesktop: store.version.downloadLatestDesktop,
-    menuButtons: buttons,
+    menuButtons: store.auth.isAuthenticated ? buttons : [],
+    isAuthenticated: store.auth.isAuthenticated,
```
seniorquico, line 53: Why was the `isAuthenticated` prop added? It doesn't appear to be used anywhere.

seniorquico (approved):

PR #591, opened 2020-12-17
Ensure checkDesktopVersion is called on native app

scottcinatl:
Runs checkDesktopVersion API call if latestDesktopVersion observable is ever undefined.

### DIFF
--- a/packages/web-app/src/modules/versions/VersionStore.ts
+++ b/packages/web-app/src/modules/versions/VersionStore.ts
@@ -1,6 +1,6 @@
 import { AxiosInstance } from 'axios'
 import compareVersions from 'compare-versions'
-import { action, autorun, flow, observable, runInAction } from 'mobx'
+import { action, flow, observable, reaction, runInAction } from 'mobx'
 import { config } from '../../config'
 import { getItem, removeItem, setItem } from '../../Storage'
 import { RootStore } from '../../Store'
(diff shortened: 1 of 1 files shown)


packages/web-app/src/modules/versions/VersionStore.ts
```diff
@@ -45,13 +45,13 @@ export class VersionStore {
       removeItem(VERSION_RELOAD_DATA)
     }
 
-    if (this.store.native.isNative) {
-      this.checkDesktopVersion()
-    }
-
     autorun(() => {
+      if (this.latestDesktopVersion === undefined) {
```
seniorquico, line 49: Should we only make the API request if we know we're in the desktop app?

```suggestion
      if (this.latestDesktopVersion === undefined && this.store.native.isNative) {
```

If yes, may want to refactor this since the next block also appears to require being in the desktop app. Even better-- could we skip creating the `autorun` if we're not in the desktop app? One less function being called!

seniorquico (changes requested):

packages/web-app/src/modules/versions/VersionStore.ts
```diff
@@ -49,15 +49,18 @@ export class VersionStore {
...
+      (latestDesktopVersion, reaction) => {
+        if (latestDesktopVersion !== undefined && this.store.native.desktopVersion) {
+          let onLatest = compareVersions.compare(this.store.native.desktopVersion, latestDesktopVersion, '>=')
+          runInAction(() => {
+            this.onLatestDesktop = onLatest
+          })
+          reaction.dispose()
+        }
+      },
+    )
```
seniorquico, line 63: So the biggest difference I can see between `autorun` & `reaction` for this use case is that `reaction` won't run once when initialized. This ensures the callback is deferred, waiting for the API response. But I think you also need to react to the `this.store.native.desktopVersion` to eliminate the race condition in these two async initializers. Haven't tested, but maybe something like this?

```ts
    if (this.store.native.isNative) {
      reaction(
        () => ({
          currentDesktopVersion: this.store.native.desktopVersion,
          latestDesktopVersion: this.latestDesktopVersion,
        }),
        ({ currentDesktopVersion, latestDesktopVersion }, reaction) => {
          if (currentDesktopVersion !== undefined && latestDesktopVersion !== undefined) {
            const onLatest = compareVersions.compare(currentDesktopVersion, latestDesktopVersion, '>=')
            runInAction(() => {
              this.onLatestDesktop = onLatest
            })
            reaction.dispose()
          }
        },
      )

      this.checkDesktopVersion()
    }
```

seniorquico (changes requested):

seniorquico (approved):

PR #595, opened 2020-12-21
Added MinecraftUsername component to Profile page

scottcinatl:
Adds form that allows user to provide a Minecraft Username associated with their Salad account


packages/web-app/src/modules/reward/RewardStore.ts
```diff
@@ -297,6 +299,15 @@ export class RewardStore {
             autoClose: false,
             onClick: () => this.store.routing.push('/account/reward-vault'),
           })
+          // TODO: Change to correct minecraft error conditional once API is created
+        } else if (error.message === 'Minecraft Username not available') {
```
seniorquico, line 303: Shouldn't this be checking for a `redemptions:requires:minecraftUsername` problem?

seniorquico (changes requested):

packages/web-app/src/modules/reward/RewardStore.ts
scottcinatl: I am not sure if this is the best way to go about this, but here is my [commit](https://github.com/SaladTechnologies/salad-applications/pull/595/commits/b6b2c921a1263a76f4a6fdfd20ab2d32a7bfcdf6) of what I have implemented so far. Might need more refactoring?

packages/web-app/src/axiosFactory.ts
```diff
@@ -4,6 +4,8 @@ import isRetryAllowed from 'is-retry-allowed'
 import { config } from './config'
 import { AuthStore } from './modules/auth'
 
+export const REQUIRES_MINECRAFT_USERNAME = 'Redemption requires a Minecraft Username.'
```
seniorquico, line 7: This should be inlined.

packages/web-app/src/modules/account-views/account-views/components/Account.tsx
```diff
@@ -27,6 +28,8 @@ class _Account extends Component<Props> {
         <Head title="Account" />
         <div className={classnames(classes.container)}>
           <EditUsername profile={profile} onSend={onSend} sending={sending} />
+          <Divider />
+          <MinecraftUsernameContainer />
```
seniorquico, line 32: Just curious... why not `EditMinecraftUsername` to match the field above?

packages/web-app/src/modules/account-views/account-views/MinecraftUsernameContainer.ts
```diff
@@ -0,0 +1,14 @@
...
+const mapStoreToProps = (store: RootStore): any => {
+  const username = store.profile.currentProfile?.minecraftUsername
+  return {
+    username: username,
+    isUpdating: store.profile.isUpdating,
+    onUpdateUsername: store.profile.updateMinecraftUsername,
+  }
+}
+
+export const MinecraftUsernameContainer = connect(mapStoreToProps, MinecraftUsername)
```
seniorquico, line 14: Is this the desired pattern, that each field in the form be connected to the store? Why isn't the overall form being connected, and the fields just acting as simple components with basic props?

packages/web-app/src/modules/account-views/account-views/components/MinecraftUsername.tsx
```diff
@@ -0,0 +1,127 @@
...
+    return errors
+  }
+
+  render() {
+    const { isUpdating, classes, username } = this.props
+
+    return (
+      <div className={classes.container}>
+        <Username>Minecraft Username</Username>
+        <P>Add your Minecraft username in order to redeem some sweet Minecraft skins!</P>
```
seniorquico, line 58: Is this the desired text? First, "skins" aren't really a Minecraft thing... they're "capes". Second, I think we're going to support subscriptions to private servers & loot boxes. Double check on the copy.

packages/web-app/src/modules/profile/ProfileStore.ts
```diff
@@ -76,4 +76,27 @@ export class ProfileStore {
...
   })
+
+  @action.bound
+  updateMinecraftUsername = flow(function* (this: ProfileStore, minecraftUsername: string) {
+    if (this.currentProfile === undefined) return
+
+    this.isUpdating = true
+
+    try {
+      let patch = yield this.axios.patch('/api/v1/profile', { minecraftUsername: minecraftUsername })
```
seniorquico, line 87: See the note about the API change on the Notion card. Easy fix:

```suggestion
      let patch = yield this.axios.patch('/api/v1/profile', { extensions: { minecraftUsername }})
```

packages/web-app/src/modules/profile/ProfileStore.ts
```diff
@@ -76,4 +76,27 @@ export class ProfileStore {
       // this.store.routing.replace('/')
     }
   })
+
+  @action.bound
+  updateMinecraftUsername = flow(function* (this: ProfileStore, minecraftUsername: string) {
+    if (this.currentProfile === undefined) return
```
seniorquico, line 82: nit:

```suggestion
    if (this.currentProfile === undefined) {
      return
    }
```

packages/web-app/src/modules/profile/models/Profile.ts
```diff
@@ -2,6 +2,7 @@ export interface Profile {
   id: string
   username: string
   email: string
+  minecraftUsername?: string
```
seniorquico, line 5:
```suggestion
  extensions?: {
    minecraftUsername?: string
  }
```

packages/web-app/src/modules/reward-views/RewardDetailsContainer.tsx
```diff
@@ -15,6 +15,15 @@ const mapStoreToProps = (store: RootStore, props: RouteComponentProps<{ id: stri
   isInCart: store.rewards.isInChoppingCart(props.match.params.id),
   onAddToCart: store.rewards.addToChoppingCart,
   onRemoveFromCart: store.rewards.removeFromChoppingCart,
+  hasMinecraftUsername: store.profile.currentProfile?.minecraftUsername !== undefined,
```
seniorquico, line 18:
```suggestion
  hasMinecraftUsername: store.profile.currentProfile?.extensions?.minecraftUsername !== undefined,
```

packages/web-app/src/modules/reward-views/components/RewardHeaderBar.tsx
```diff
@@ -92,9 +92,11 @@ interface Props extends WithStyles<typeof styles> {
   authenticated?: boolean
   onBack?: () => void
   onRedeem?: (reward?: Reward) => void
+  onFurtherActionRequired: () => void
   isInCart?: boolean
   onAddToCart?: (reward: Reward) => void
   onRemoveFromCart?: (reward: Reward) => void
+  hasMinecraftUsername: boolean
```
seniorquico, line 99: Is there a better way to genericize this? Otherwise, won't we end up with dozens of flags for different types of (future) requirements?

packages/web-app/src/modules/reward-views/components/RewardHeaderBar.tsx
```diff
@@ -127,6 +137,8 @@ class _RewardHeaderBar extends Component<Props> {
     const promoGame: boolean = reward ? reward?.price === 0 : false
     const hasBalance = reward ? reward?.price <= balance : false
 
+    const requiresMinecraftUsername = reward?.tags?.includes('requires-minecraft-username') && !hasMinecraftUsername
```
seniorquico, line 140: This feels like business logic, something that shouldn't be inside the view layer.

packages/web-app/src/modules/reward-views/components/RewardHeaderBar.tsx
```diff
@@ -163,10 +175,15 @@ class _RewardHeaderBar extends Component<Props> {
...
+              {requiresMinecraftUsername && authenticated && (
+                <div className={classnames(classes.priceText, classes.stockLabel, classes.insufficientBalanceLabel)}>
+                  <SmartLink to="/settings/summary">Add Minecraft Username</SmartLink>
+                </div>
+              )}
             </div>
             <Button
               className={classes.buyButton}
-              onClick={this.handleRedeem}
+              onClick={requiresMinecraftUsername ? this.furtherActionRequired : this.handleRedeem}
```
seniorquico, line 186: If `furtherActionRequired` is a generic prop, with a specific action (routing to the profile page to set the Minecraft username), then it feels like these other bits should be moved outside the component. I think this helps with preparing for different, future reward requirements.

seniorquico (changes requested):

packages/web-app/src/modules/account-views/account-views/components/Account.tsx
scottcinatl: The `EditUsername` component only allows the ability to edit the username, you can not create a username in that component. Whereas the `MinecraftUsername` component allows for creation and updating.

packages/web-app/src/modules/account-views/account-views/components/MinecraftUsername.tsx
scottcinatl: No, yeah this is placeholder text. I will add a TODO flag around this and ping Jared/Dan to get the desired copy.

packages/web-app/src/modules/profile/ProfileStore.ts
```diff
@@ -76,4 +76,33 @@ export class ProfileStore {
...
+      return
+    }
+
+    this.isUpdating = true
+
+    try {
+      let patch = yield this.axios.patch('/api/v1/profile', {
+        extensions: {
+          minecraftUsername: minecraftUsername,
+        },
```
seniorquico, line 92: The `PATCH` method is cool with merging in changes to top-level properties... but it doesn't preserve any child properties. For the most part, this is why we don't frequently use nested objects. We need to explicitly preserve any other extension properties, otherwise this will be _deleting_ the others.

```suggestion
        extensions: this.currentProfile.extensions === undefined
          ? {
              minecraftUsername: minecraftUsername,
            }
          : {
              ...this.currentProfile.extensions,
              minecraftUsername: minecraftUsername,
            },
```

seniorquico (changes requested):

seniorquico (approved):

PR #613, opened 2021-01-22
Updated Fallback Error Modal

scottcinatl:
The updated Fallback Error Modal now provides the user with an opportunity to switch to an alternative mining type while also presenting them with helpful support links.

### DIFF
--- a/packages/web-app/src/modules/error-views/components/FallbackErrorPage.tsx
+++ b/packages/web-app/src/modules/error-views/components/FallbackErrorPage.tsx
@@ -1,24 +1,71 @@
 import { Component } from 'react'
+import withStyles, { WithStyles } from 'react-jss'
 import { SmartLink } from '../../../components'
+import { Button } from '../../../components/Button'
 import { ErrorPage } from '../../../components/ErrorPage'
+import { SaladTheme } from '../../../SaladTheme'
 
-interface Props {
-  onCloseClicked?: () => void
+const styles = (theme: SaladTheme) => ({
+  buttons: {
+    display: 'flex',
+  },
+  button: {
+    color: 'red',
+    display: 'flex',
+    flex: 1,
+    justifyContent: 'center',
+  },
+  buttonText: {
+    color: theme.darkBlue,
+  },
+  link: {
+    cursor: 'pointer',
+    textDecoration: 'underline',
+  },
+  padding: {
+    paddingBottom: '1rem',
+  },
+})
+
+interface Props extends WithStyles<typeof styles> {
+  alternativeMinerType: 'GPU' | 'CPU'
+  currentMinerType: 'GPU' | 'CPU'
+  onCloseClicked: () => void
+  onOpenSupportTicket: () => void
+  onSwitchMiningType: () => void
+  gpuMiningEnabled: boolean
 }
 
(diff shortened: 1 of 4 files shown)


packages/web-app/src/modules/error-views/components/FallbackErrorPage.tsx
```diff
@@ -1,24 +1,65 @@
...
   render() {
+    const { gpuMiningEnabled, onCloseClicked, onOpenSupportTicket, onSwitchMiningType, classes } = this.props
     return (
-      <ErrorPage title="Initialization Error" onCloseClicked={this.props.onCloseClicked}>
-        <div>
-          Hey Chef, it looks like Salad can't run on your machine right now. Unfortunately, this happens sometimes
-          despite having a compatible GPU. We're working hard to resolve this, but in the meantime please submit a bug
-          report by following the link below.
+      <ErrorPage
+        title={`Salad is Unable to Run on your ${gpuMiningEnabled ? 'GPU' : 'CPU'}`}
```
seniorquico, line 42: The inline ternaries are repeated over and over... you should refactor them and only run the check once.

packages/web-app/src/modules/zendesk/Zendesk.ts
```diff
@@ -152,4 +152,37 @@ export class Zendesk {
       console.error(e)
     }
   }
+
+  openSupportTicket() {
+    if (window && window.zE) {
+      try {
+        // Don't skip the Help Center every time...
+        window.zE('webWidget:on', 'close', () => {
```
seniorquico, line 160: Is there a consequence to registering a callback every time the `openSupportTicket` function is called? Does it result in multiple subscriptions/callbacks? Should this be moved into the startup/initialization routine?

seniorquico (changes requested):

packages/web-app/src/modules/zendesk/Zendesk.ts
scottcinatl: I couldn't find any definitive answers in the documentation, but after playing around when the on close event, I believe that it can only be called/set during that close event. It doesn't seem like there is a way to set a standard on close event in the core settings for the widget.

packages/web-app/src/modules/error-views/components/FallbackErrorPage.tsx
```diff
@@ -1,24 +1,78 @@
...
+          <span className={classes.link} onClick={onOpenSupportTicket}>
+            opening a Support Ticket
+          </span>{' '}
+          or getting help from our{' '}
+          <SmartLink to="https://forum.salad.io/c/community-support/6">Support Forum</SmartLink>.
         </div>
-        <div style={{ paddingTop: '1rem' }}>
-          <SmartLink to="https://www.salad.io/support/">Salad Support</SmartLink>
+        <div className={classes.padding}>
+          In the meantime, try switching to GPU mining so that you keep earning while we work to resolve your issue.
```
seniorquico, line 51: Should this text also be dynamic based on the currently selected mode (GPU vs CPU)?

seniorquico (changes requested):

seniorquico (approved):

PR #616, opened 2021-01-27
Added Strapi Storefront Page prototype

scottcinatl:
Adds a prototype in Storybook of the existing home page while consuming mock Strapi data.


seniorquico (changes requested):

packages/web-app/src/modules/storefront-views/blocks/StorefrontHeroBlock.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+interface Props {
+  block: StorefrontHeroBlockProps
+}
+
+export const StorefrontHeroBlock = ({ block }: Props) => {
+  return (
+    <StorefrontHero key={block.id} title={block.title}>
+      {block.items.map((hero) => {
+        const heading = hero.reward?.name || hero.heading
```
seniorquico, line 14: These are flipped... We want to use the explicitly defined value first, and we want to _fallback_ to the reward (if defined).

```suggestion
        const heading = hero.heading || hero.reward?.name
```

packages/web-app/src/modules/storefront-views/blocks/StorefrontHeroBlock.tsx
```diff
@@ -0,0 +1,38 @@
...
+        const subHeading = hero.reward?.price ? `$${hero.reward?.price.toFixed(2)}` : hero.subheading
```
seniorquico, line 15: These are flipped... also, converting `hero.reward?.price` to a truthy/falsy value doesn't work when the price is free. For consistency, shouldn't a free price be transformed into a "FREE" string like in the reward list?

```suggestion
        const subHeading = hero.subheading || `$${hero.reward?.price.toFixed(2)}`
```

packages/web-app/src/modules/storefront-views/blocks/StorefrontHeroBlock.tsx
```diff
@@ -0,0 +1,38 @@
...
+        const buttonLink = hero.button_link ? hero.button_link : rewardRoute(hero.reward?.uuid)
```
seniorquico, line 16: nit:

```suggestion
        const buttonLink = hero.button_link || rewardRoute(hero.reward?.uuid)
```

packages/web-app/src/modules/storefront-views/blocks/StorefrontHeroBlock.tsx
```diff
@@ -0,0 +1,38 @@
...
+        const image = hero.reward?.image || `http://localhost:1337${hero.image.url}` // Change strapi api url to env variable when connecting
```
seniorquico, line 17: These are flipped... but also, please move the Strapi API origin into the `.env` file(s) and use the `URL` constructor to normalize the value. For example:

```ts
new URL(hero.image.url, 'http://localhost:1337')
```

The MDN docs note this works as expected, using the second paramter as a fallback:

> If `url` is a relative URL, `base` is required, and will be used as the base URL. If `url` is an absolute URL, a given `base` will be ignored.

packages/web-app/src/modules/storefront-views/blocks/StorefrontRewardBlock.tsx
```diff
@@ -0,0 +1,40 @@
...
+export const StorefrontRewardBlock = ({ block, onClick }: Props) => {
+  return (
+    <Fragment key={block.id}>
+      <StorefrontRewardSlider
+        key={block.id}
+        title={block.title}
+        viewAllRoute={block.buttons[0].link}
+        viewAllTitle={block.buttons[0].label}
+      >
+        {block.rewards.slice(0, 20).map((reward) => {
```
seniorquico, line 20: We should not be limiting this to the first 20 rewards.

packages/web-app/src/modules/storefront-views/blocks/StorefrontHeroBlock.tsx
```diff
@@ -0,0 +1,38 @@
...
+        const outOfStock = hero.reward?.quantity === 0
+        const lowQuanity = hero.reward?.quantity !== undefined && hero.reward.quantity > 0
+        return (
+          <StorefrontHeroItem
+            key={hero.id}
+            altText={hero.image.alternativeText || ''}
+            body={hero.body}
+            buttonLabel={hero.button_label}
+            buttonLink={buttonLink}
```
seniorquico, line 26: Shouldn't click actions follow the "standard" callback pattern? If no... why does this one pass a link down whereas the reward list passes a callback down? It seems weird to apply different.

packages/web-app/src/modules/storefront-views/blocks/StorefrontRewardBlock.tsx
```diff
@@ -0,0 +1,40 @@
...
+  onClick: (id: string) => void
+}
+
+export const StorefrontRewardBlock = ({ block, onClick }: Props) => {
+  return (
+    <Fragment key={block.id}>
+      <StorefrontRewardSlider
+        key={block.id}
+        title={block.title}
+        viewAllRoute={block.buttons[0].link}
```
seniorquico, line 17: What if no buttons are defined?

packages/web-app/src/modules/storefront/models/StorefrontRewardItemProps.ts
```diff
@@ -0,0 +1,11 @@
+export interface StorefrontRewardItemProps {
+  id: string
+  name: string
+  uuid: string
+  price: number
+  inStock: boolean
+  quantity: number
+  created_at: string
+  updated_at: string
+  image: string
```
seniorquico, line 10:
```suggestion
  coverImage: string
  heroImage: string
```

seniorquico (changes requested):

packages/web-app/src/modules/storefront-views/blocks/StorefrontHeroBlock.tsx
scottcinatl: Changed the `RewardItem` to match the `HeroItem` where each Reward is wrapped by our SmartLink component. This allows to take advantage of the ability of the `SmartLink` component to add tracking metrics if needed.

packages/web-app/src/modules/storefront-views/blocks/StorefrontContentBlock.tsx
```diff
@@ -0,0 +1,18 @@
+import { StorefrontContentBlockProps } from '../../storefront/models'
+import { StorefrontContentPanel } from '../components/StorefrontContentPanel'
+
+interface Props {
+  block: StorefrontContentBlockProps
+}
+
+export const StorefrontContentBlock = ({ block }: Props) => {
```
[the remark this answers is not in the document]
scottcinatl: Consistency with the other block components, also if content blocks fields evolve to be more dynamic with the different banners shown in the Notion resource page (coulldn't find the link, but it the only with all possible designs of what a banner could be), the higher order block component could manage that decision logic what component renders.

packages/web-app/src/modules/storefront/models/StorefrontBlockComponents.ts
```diff
@@ -0,0 +1,32 @@
+import { StorefrontHeroItemProps, StorefrontImageProps, StorefrontRewardItemProps } from './index'
+
+export interface StrapiBlock {
+  __component: string
```
seniorquico, line 4: In the future, leverage TypeScript's ["discriminated union" feature](https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html#discriminating-unions). It simplifies narrowing polymorphic types.

packages/web-app/src/modules/storefront/models/StorefrontHeroItemProps.ts
```diff
@@ -0,0 +1,64 @@
...
+        name: string
+        hash: string
+        ext: '.png' | '.jpg' | '.png'
+        mime: string
+        width: number
+        height: number
+        size: number
+        path?: string
+        url: string
+      }
```
seniorquico, line 29: Looks like a repeated element. Probably good for a refactor into a reusable type.

seniorquico (approved):

PR #618, opened 2021-01-28
Mixpanel enhancements

scottcinatl:
Improves/Adds Mixpanel tracking events
- Refactors currentPath argument for link tracking events
- Adds category data and improve naming of color type for Toast Notification tracking events
- Adds tracking for error pages viewed by the user and buttons clicked

### DIFF
--- a/packages/web-app/src/modules/error-views/AntiVirusErrorContainer.ts
+++ b/packages/web-app/src/modules/error-views/AntiVirusErrorContainer.ts
@@ -2,14 +2,19 @@ import { connect } from '../../connect'
 import { RootStore } from '../../Store'
 import { AntiVirusErrorPage } from './components/AntiVirusErrorPage'
 
-const mapStoreToProps = (store: RootStore): any => ({
-  onCloseClicked: () => store.ui.hideModal(),
-  avLinks: [
-    {
-      url: 'https://support.salad.io/hc/en-us/sections/360008458292-Anti-Virus',
-      text: 'Anti-Virus Configuration Guides',
-    },
-  ],
-})
+const mapStoreToProps = (store: RootStore): any => {
+  return {
+    onCloseClicked: () => store.ui.hideModal(true),
+    avLinks: [
+      {
+        url: 'https://support.salad.io/hc/en-us/sections/360008458292-Anti-Virus',
+        text: 'Anti-Virus Configuration Guides',
+        trackingInfo: {
+          label: 'Anti-Virus Configuration Guides',
+        },
+      },
+    ],
+  }
+}
 
 export const AntiVirusErrorContainer = connect(mapStoreToProps, AntiVirusErrorPage)
--- a/packages/web-app/src/modules/error-views/CudaErrorContainer.ts
+++ b/packages/web-app/src/modules/error-views/CudaErrorContainer.ts
@@ -3,7 +3,7 @@ import { RootStore } from '../../Store'
 import { CudaErrorPage } from './components/CudaErrorPage'
 
 const mapStoreToProps = (store: RootStore): any => ({
-  onCloseClicked: () => store.ui.hideModal(),
+  onCloseClicked: () => store.ui.hideModal(true),
 })
 
 export const CudaErrorContainer = connect(mapStoreToProps, CudaErrorPage)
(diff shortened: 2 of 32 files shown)


packages/web-app/src/modules/error-views/AntiVirusErrorContainer.ts
```diff
@@ -2,14 +2,22 @@ import { connect } from '../../connect'
...
-    {
-      url: 'https://support.salad.io/hc/en-us/sections/360008458292-Anti-Virus',
-      text: 'Anti-Virus Configuration Guides',
-    },
-  ],
-})
+const mapStoreToProps = (store: RootStore): any => {
+  if (!store.saladBowl.hasViewedAVErrorPage) {
+    store.analytics.trackErrorPageViewed('Anti-Virus Error')
+  }
```
seniorquico, line 8: We already have an event for when this dialog is shown: `Mining Error`. Additionally, components should be relatively dumb, just passing data from and firing events to a store. This business logic should be kept within `SaladBowlStore`.

packages/web-app/src/modules/error-views/components/UnknownErrorPage.tsx
```diff
@@ -15,8 +15,15 @@ export class UnknownErrorPage extends Component<Props> {
         <div>
           We are sorry for the inconvenience, Salad doesn't seem to be working properly. If this issue persists,
           something else may be causing the problem. If so, please contact
-          <SmartLink to="https://www.salad.io/support/">Salad Support</SmartLink>, join the
-          <SmartLink to="https://discord.gg/XzyRcd8"> Discord </SmartLink> for updates and send us your logs.
+          <SmartLink to="https://www.salad.io/support/" trackingInfo={{ label: 'Salad Support' }}>
+            Salad Support
+          </SmartLink>
```
seniorquico, line 20: I'm not in love with this pattern... it looks like we're duplicating content now. Why not just stringify the `children`?

packages/web-app/src/modules/salad-bowl/SaladBowlStore.ts
```diff
@@ -51,6 +52,9 @@ export class SaladBowlStore implements IPersistentStore {
   @observable
   public gpuMiningOverridden: boolean = false
 
+  @observable
+  public hasViewedAVErrorPage: boolean = false
```
seniorquico, line 56: Should be `private`, and likely doesn't need to be `observable`. What subscriber is reacting to this?

packages/web-app/src/modules/salad-bowl/SaladBowlStore.ts
```diff
@@ -283,6 +288,8 @@ export class SaladBowlStore implements IPersistentStore {
 
   @action.bound
   start = flow(function* (this: SaladBowlStore, reason: StartReason, startTimestamp?: Date, choppingTime?: number) {
+    this.hasViewedAVErrorPage = false
```
seniorquico, line 291: This should probably move below the guard to ensure we don't accidentally reset it when a new session hasn't actually been started. i.e. below this block:

```ts
if (this.isRunning) {
  return
}
```

seniorquico (changes requested):

packages/web-app/src/modules/error-views/CudaErrorContainer.ts
```diff
@@ -2,8 +2,12 @@ import { connect } from '../../connect'
...
-const mapStoreToProps = (store: RootStore): any => ({
-  onCloseClicked: () => store.ui.hideModal(),
-})
+const mapStoreToProps = (store: RootStore): any => {
+  store.analytics.trackErrorPageViewed('Cuda Error')
+
+  return {
+    onCloseClicked: () => store.ui.hideModal(),
+  }
+}
```
seniorquico, line 11:
```suggestion
const mapStoreToProps = (store: RootStore): any => ({
  onCloseClicked: () => {
    // TODO: Refactor business logic like this into a store function (e.g. `hideModal`)
    store.analytics.trackErrorPageViewed('Cuda Error')
    store.ui.hideModal()
  },
})
```

seniorquico (changes requested):

seniorquico (approved):

PR #621, opened 2021-02-02
Prevents unauthenticated user from seeing compatibility issue modal

scottcinatl:
Checks if user is authenticated before showing compatibility issue on Start Button click.

### DIFF
--- a/packages/web-app/src/modules/salad-bowl/SaladBowlStore.ts
+++ b/packages/web-app/src/modules/salad-bowl/SaladBowlStore.ts
@@ -8,7 +8,7 @@ import { NotificationMessageCategory } from '../notifications/models'
 import { IPersistentStore } from '../versions/IPersistentStore'
 import { getPluginDefinitions } from './definitions'
 import { Accounts, BEAM_WALLET_ADDRESS, ETH_WALLET_ADDRESS, getNiceHashMiningAddress } from './definitions/accounts'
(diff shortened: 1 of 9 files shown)


seniorquico (changes requested):
- Combine `onStartButtonClicked` and `toggleRunning`.
- Audit `start` calls to ensure the appropriate dialog is always shown.
- Remove dead `toggleRunning` call in `MainTitlebarContainer`.

packages/web-app/src/modules/salad-bowl/SaladBowlStore.ts
```diff
@@ -280,11 +290,27 @@ export class SaladBowlStore implements IPersistentStore {
   }
 
   @action
-  toggleRunning = () => {
+  toggleRunning = (startReason: StartReason, startType: StartActionType) => {
```
seniorquico, line 293: I'm concerned that some calls may end up inadvertently _stopping_ the miner when the intent was to start it. For example, what happens if the auto-start logic misfires and calls `toggleRunning` when we've already started? I think it would result in calling `stop`. I think we should be explicit our actions, and maybe it's as simple as renaming `toggleRunning` to something else that doesn't imply it's toggling anything.

seniorquico (changes requested):

seniorquico (approved):

PR #625, opened 2021-02-11
Added new AntiVirus Guides to Storybook

scottcinatl:
1. Adds two new Error Pages to the app that provides guides from our Zendesk help center for users to follow/troubleshoot their Anti-Virus issues.
2. Adds fix to Storybook production error


packages/web-app/src/modules/error-views/components/FallthroughAntiVirusErrorPage.tsx
```diff
@@ -0,0 +1,225 @@
...
+    fontFamily: theme.fontGroteskLight25,
+    justifyContent: 'center',
+    lineHeight: '150%',
+  },
+  subtitle: {
+    fontSize: theme.medium,
+  },
+  supportReminder: {
+    fontSize: theme.small,
+    position: 'absolute',
```
seniorquico, line 92: Is there any way to make this responsive to the actual position of the widget? Other than us changing it's position in the Zendesk config, I wonder if the absolute positioning could also fail with different edge cases such as page zooming, non-default DPIs, and broken fonts?

packages/web-app/src/modules/error-views/components/FallthroughAntiVirusErrorPage.tsx
```diff
@@ -0,0 +1,225 @@
...
+  onCloseClicked?: () => void
+  onViewArticle?: (id: number) => void
+  onViewAVList: () => void
+}
+
+interface State {
+  webWidgetShowing: boolean
+}
+
+const tooltipText =
```
seniorquico, line 121: This should be inline in the JSX.

packages/web-app/src/modules/error-views/components/FallthroughAntiVirusErrorPage.tsx
```diff
@@ -0,0 +1,225 @@
...
+class _FallthroughAntiVirusErrorPage extends Component<Props, State> {
+  constructor(props: Props) {
+    super(props)
+    this.state = {
+      webWidgetShowing: false,
+    }
+  }
+
+  componentDidMount() {
+    const webWidgetInterval = setInterval(function () {
```
seniorquico, line 133: Should this continue to run every second as an interval? Or should this be a one-time check using `setTimeout`?

packages/web-app/src/modules/error-views/components/FallthroughAntiVirusErrorPage.tsx
```diff
@@ -0,0 +1,225 @@
...
+      getWidgetDisplay()
```
seniorquico, line 134: This isn't cancelled when the component unmounts, and it may result in a `setState` call on an unmounted component.

packages/web-app/src/modules/error-views/components/FallthroughAntiVirusErrorPage.tsx
```diff
@@ -0,0 +1,225 @@
...
+                  Your Anti-Virus has blocked all of our miners. so you'll need to whitelist Salad with your Anti-Virus
+                  in order to get chopping. Can't whitelist? Earn anyway using{' '}
+                  <SmartLink to="/earn/offerwall">Offerwalls</SmartLink>, or reach out to the{' '}
+                  <SmartLink to="https://forum.salad.io/">support forum</SmartLink> for help!
+                </p>
+              )}
+            </div>
+          </div>
+          {article ? (
+            <ScrollableContent width={650}>
```
seniorquico, line 187: How does this fixed width respond to different window sizes? Why are we constraining it to this narrow width? Was this picked based on some other, existing layout size?

packages/web-app/src/modules/error-views/components/FallthroughAntiVirusErrorPage.tsx
```diff
@@ -0,0 +1,225 @@
...
+              <div className={classes.content} dangerouslySetInnerHTML={{ __html: article }} />
+            </ScrollableContent>
+          ) : articleList ? (
+            <ScrollableContent width={650}>
+              <ul className={classes.list}>
+                {articleList.map((article) => (
+                  <li
+                    key={article.id}
+                    className={classes.listItem}
+                    onClick={onViewArticle ? () => onViewArticle(article.id) : undefined}
```
seniorquico, line 197: Why would `onViewArticle` be `undefined`? What can the user do if it's `undefined`?

packages/web-app/src/modules/error-views/components/FallthroughAntiVirusErrorPage.tsx
```diff
@@ -0,0 +1,225 @@
...
+                  >
+                    {article.name}
+                  </li>
+                ))}
+              </ul>
+            </ScrollableContent>
+          ) : null}
+          {antivirus && (
+            <div className={classes.bottomMessage}>
+              Use a different Anti-Virus provided?{' '}
```
seniorquico, line 207: > Use a different Anti-Virus provided?

This doesn't sound correct. Should it be:

> Use a different Anti-Virus product?

Also, is it correct to capitalize anti-virus this way?

packages/web-app/src/modules/error-views/components/FallthroughAntiVirusErrorPage.tsx
```diff
@@ -0,0 +1,225 @@
...
+              <span className={classes.link} onClick={onViewAVList}>
+                Select it from this list
+              </span>
+              .
+            </div>
+          )}
+        </div>
+        {webWidgetShowing && (
```
seniorquico, line 215: What about a smoother animation to transition this in? It just feels odd to flash in a second or two after the page has already finished rendering.

seniorquico (changes requested):

packages/web-app/src/modules/error-views/components/FallthroughAntiVirusErrorPage.tsx
```diff
@@ -0,0 +1,225 @@
...
+    marginBottom: 10,
+    textDecoration: 'underline',
+  },
+  page: {
+    position: 'fixed',
+    top: 0,
+    bottom: 0,
+    left: 0,
+    right: 0,
+    zIndex: 3002,
```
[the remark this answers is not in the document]
scottcinatl: When I was using the other overlay component, I was running into issues with the zIndex of our current Tooltip health icon. 
Capture.

So either way, I was having to change a zIndex somewhere. Maybe we could create a task to better handle our zIndex usage?

packages/web-app/src/modules/error-views/components/FallthroughAntiVirusErrorPage.tsx
```diff
@@ -0,0 +1,225 @@
...
+  bottomMessage: {
+    fontSize: theme.medium,
+    margin: '25px 0',
+    textAlign: 'center',
+  },
+  checkBox: {
+    display: 'flex',
+    justifyContent: 'center',
+  },
+  closeButton: {
```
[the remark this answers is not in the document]
scottcinatl: Not sure these will be used in the future, but will flag it in codemark.

packages/web-app/src/modules/error-views/components/FallthroughAntiVirusErrorPage.tsx
scottcinatl: The size of the content was based off Figma design for this error page.

packages/web-app/src/components/ScrollableContent.tsx
```diff
@@ -0,0 +1,48 @@
+import { Component } from 'react'
+import Scrollbars from 'react-custom-scrollbars'
```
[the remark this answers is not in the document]
scottcinatl: Yes, this is already a dependency. We have a Scrollbars component that handles scroll for a whole page, whereas the ScrollableContent component I created creates a scrollable contained box.

packages/web-app/src/modules/error-views/components/FallthroughAntiVirusErrorPage.tsx
scottcinatl: I couldn't find any clean way in the Zendesk WebWidget API to determine where the position of the widget is.

packages/web-app/src/modules/error-views/components/FallthroughAntiVirusErrorPage.tsx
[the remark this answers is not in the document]
seniorquico: Ok. Yes, we need to tackle this better. Explicitly managing z-indexes is always feels like a nightmare, has to be done globally across the app. Let's spin this out as a refactor task in the backlog, and it'll probably be something to include in the scope of the component library project.

packages/web-app/src/modules/error-views/components/NewAntiVirusErrorPage.tsx
```diff
@@ -0,0 +1,257 @@
...
+              Psst! This information can also be found by clicking the support button and searching for 'anti-virus'.
+            </div>
+          )}
+        </div>
+      </ModalPage>
+    )
+  }
+}
+
+export const AntiVirusErrorPage = withStyles(styles)(_AntiVirusErrorPage)
```
seniorquico, line 257: The file name should be `AntiVirusErrorPage`, to match the component name.

seniorquico (changes requested):

seniorquico (approved):

PR #630, opened 2021-02-20
Added messaging to notify users of new authentication flow

scottcinatl:


packages/web-app/src/modules/auth-views/components/FramePage.tsx
```diff
@@ -86,24 +101,33 @@ export const FramePage = withStyles(styles)(
...
+              src={frameUrl}
+              title={frameTitle}
             />
-          )}
+            <NotMobile>
+              {frameLoaded && (
+                <P className={classes.subtitle}>
+                  Each time you sign in to Salad, you’ll receive a fresh <br /> one-time code in your email. Using these
+                  one-time codes helps <br /> protect your Salad balance and rewards.
+                </P>
```
seniorquico, line 124: This should be responsive. Instead of line breaks, the box should have a set width. The text will automatically flow onto multiple lines as needed.

seniorquico (approved):

PR #636, opened 2021-02-25
Connected new Start Button and Don't Lose Progress page

scottcinatl:
1. Adds new Start Button component which showcases the progress of a user's machine prepping before chopping.
2. Shows Don't Lose Your Progress page to a user who clicks the Stop button while still prepping their machine.


packages/web-app/src/UIStore.ts
```diff
@@ -12,7 +14,26 @@ export enum ErrorPageType {
 }
 
 export class UIStore {
-  constructor(private readonly store: RootStore) {}
+  constructor(private readonly store: RootStore) {
+    autorun(() => {
```
seniorquico, line 18: No.

seniorquico (changes requested):

seniorquico (approved):

PR #640, opened 2021-03-03
Connect storefront

scottcinatl:
1. Connects new storefront page to App that pulls data from Strapi


packages/web-app/src/modules/storefront/StorefrontStore.ts
```diff
@@ -0,0 +1,99 @@
...
+  StorefrontHeroBlockProps,
+  StorefrontHeroItemProps,
+  StorefrontPageProps,
+  StorefrontRewardBlockProps,
+  StorefrontRewardItemProps,
+} from './models'
+import { StorefrontResource } from './models/StorefrontResource'
+
+export class StorefrontStore {
+  private rewards: Record<string, StorefrontRewardItemProps> = {}
```
seniorquico, line 16: We should be sync'ing up with the `RewardStore`.

seniorquico (changes requested):

seniorquico (approved):

seniorquico (approved):

PR #643, opened 2021-03-10
Added New Reward Vault List UI to Storybook

scottcinatl:
Along with new Storybook UI components, this updates the Reward Item model in the Reward Store to reflect the addition of the status property.

### DIFF
--- a/packages/web-app/src/modules/vault-views/components/NewVaultList.tsx
+++ b/packages/web-app/src/modules/vault-views/components/NewVaultList.tsx
@@ -218,9 +218,9 @@ class _NewVaultList extends Component<Props, State> {
 
                 {redemptions.map((reward) => {
                   const { id, name, price, timestamp, code, status } = reward
-                  const incompleteItem = status === RewardVaultStatus.CANCELLED || status === RewardVaultStatus.PENDING
+                  const incompleteItem = status === RewardVaultStatus.CANCELED || status === RewardVaultStatus.PENDING
                   const isPending = status === RewardVaultStatus.PENDING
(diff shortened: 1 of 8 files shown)


packages/web-app/src/modules/vault/models/RewardVaultStatus.ts
```diff
@@ -0,0 +1,5 @@
+export enum RewardVaultStatus {
+  CANCELLED = 'cancelled',
+  COMPLETED = 'completed',
+  PENDING = 'pending',
+}
```
seniorquico, line 5: I assume that this is the client-side representation of our API model. If so, it must be in-sync with our API model:

```cs
    public enum RedemptionStatus
    {
        Unknown,
        Created,
        Failed,
        Complete
    }
```

Notably, the `canceled` value should be `failed`. This will be the internal value used in the code... but we can/should still _present_ that status as "Canceled" to the end user.

seniorquico (changes requested):

packages/web-app/src/modules/vault-views/components/NewVaultList.tsx
```diff
@@ -0,0 +1,452 @@
...
+  }
+
+  componentDidMount = () => {
+    const { startRefresh } = this.props
+
+    this.sortRedemptionsByRedeemedDate(true)
+
+    this.setState({
+      redemptions: this.props.redemptions,
+    })
```
seniorquico, line 315: 🛑

seniorquico (changes requested):

seniorquico (approved):

PR #645, opened 2021-03-11
Replaced old home page with Strapi Storefront page

scottcinatl:
Along with replacing the old home page, this PR also includes minor UI tweaks to the storefront page: 
- Add Margin to Top Chop Carousel
- Remove link to Reward List Title
- Add logic to guard against no Reward List buttons 
- Fix the incorrect showing of the skeleton reward items when the storefront page fetches new data after initial load


packages/web-app/src/modules/storefront-views/components/StorefrontHero.tsx
```diff
@@ -9,6 +9,7 @@ import { StorefrontHeroButtonGroup } from './StorefrontHeroButtonGroup'
 const styles = (theme: SaladTheme) => ({
   container: {
     color: theme.lightGreen,
+    marginBottom: '60px',
```
seniorquico, line 12: We only need/want the margin when it's a carousel block immediately followed by a content block. The padding shouldn't be there when it's followed by a reward list block.

seniorquico (changes requested):

seniorquico (approved):

PR #648, opened 2021-03-15
Updated AV Error Page UI

scottcinatl:


packages/web-app/src/modules/error-views/components/AntiVirusScrollbar.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+const rightPosition = 250
+
+const styles = (theme: SaladTheme) => ({
+  scrollThumb: {
+    backgroundColor: `${theme.lightGreen}`,
+    border: `solid 1px ${theme.lightGreen}`,
+    cursor: 'pointer',
+    borderRadius: 3,
+    zIndex: 3001,
```
seniorquico, line 14: Why can't we do this without a z-index? What would it take to do it without one? There has to be some way to do it with normal ordering.

seniorquico (changes requested):

seniorquico (approved):

PR #651, opened 2021-03-16
Added async redemption error flow

scottcinatl:


packages/web-app/src/modules/reward/RewardStore.ts
```diff
@@ -241,6 +251,37 @@ export class RewardStore {
             onClick: () => this.store.routing.push('/settings/summary'),
             type: 'error',
           })
+        } else if (error.message === REDEMPTIONS_INVALID_PRICE) {
+          this.clearRedemptionId()
+          this.loadReward(reward.id)
+
+          this.store.notifications.sendNotification({
+            category: NotificationMessageCategory.Error,
+            title: 'The reward price is incorrect.',
```
scottcinatl, line 260: @seniorquico The copy for these error messages is being created by Keith, so these will change before being merged in.

packages/web-app/src/modules/reward/RewardStore.ts
```diff
@@ -221,18 +242,7 @@ export class RewardStore {
...
-          //Show an order processing notification
-          this.store.notifications.sendNotification({
-            category: NotificationMessageCategory.Redemption,
-            title: `Thank you for ordering ${reward.name}!`,
-            message: 'Congrats on your pick! Your item is on its way. Check your reward vault for more details.',
-            autoClose: false,
-            onClick: () => this.store.routing.push('/account/reward-vault'),
-          })
-        } else if (error.message === REQUIRES_MINECRAFT_USERNAME) {
+        if (error.message === REQUIRES_MINECRAFT_USERNAME) {
```
seniorquico, line 245: We need to move away from messages to the unique codes. In the future, the messages may be localized.

packages/web-app/src/modules/reward/RewardStore.ts
```diff
@@ -241,6 +251,37 @@ export class RewardStore {
...
+          this.clearRedemptionId()
+          this.store.notifications.sendNotification({
+            category: NotificationMessageCategory.Error,
+            title: 'Sorry, this reward no longer exists.',
+            message: 'Go to our storefront page to view all our other great rewards!',
+            autoClose: false,
+            onClick: () => this.store.routing.push('/'),
+            type: 'error',
+          })
+        } else if (error.status === 409 || error.response?.status === 409) {
```
seniorquico, line 276: Why the different pattern? Should the same be applied to the checks above? Seems odd that only the `409` status code would require this, no?

packages/web-app/src/axiosFactory.ts
```diff
@@ -5,6 +5,7 @@ import { config } from './config'
 import { AuthStore } from './modules/auth'
 
 export const REQUIRES_MINECRAFT_USERNAME = 'Redemption requires a Minecraft Username.'
+export const REDEMPTIONS_INVALID_PRICE = 'The client reward price is out of sync.'
```
seniorquico, line 8: These error messages seem superfluous. Are they being used anywhere? If you really need to pipe through a message on the error object, the problem details response contains a message property.

packages/web-app/src/modules/reward/RewardStore.ts
```diff
@@ -1,6 +1,7 @@
 import { AxiosInstance, AxiosResponse } from 'axios'
 import { action, computed, flow, observable } from 'mobx'
-import { REQUIRES_MINECRAFT_USERNAME } from '../../axiosFactory'
+import { v4 as uuidv4 } from 'uuid'
+import { REDEMPTIONS_INVALID_PRICE, REQUIRES_MINECRAFT_USERNAME } from '../../axiosFactory'
```
seniorquico, line 4: This doesn't feel right. The axios client shouldn't have business logic elements. The error type codes are redemption-specific, and they should be living in this module.

packages/web-app/src/modules/reward/RewardStore.ts
```diff
@@ -159,6 +167,10 @@ export class RewardStore {
       return
     }
 
+    if (this.redemptionId === undefined) {
```
seniorquico, line 170: What if the last redemption _attempt_ was for a different reward? Should that automatically trigger creating a new redemption ID?

seniorquico (changes requested):

seniorquico (approved):

PR #652, opened 2021-03-18
Throttles AV Error Page to only show once per session

scottcinatl:
1. Throttles AV Error Page to only show once
2. Updates scrollbar and icon design on AV Page
2. Updates content Don't Lose Progress and Fallback Pages


packages/web-app/src/modules/salad-bowl/SaladBowlStore.ts
```diff
@@ -567,4 +562,9 @@ export class SaladBowlStore implements IPersistentStore {
     //Saves the new value locally so it will automatically be loaded next time
     Storage.setItem(CPU_MINING_OVERRIDDEN, value)
   }
+
+  @action
+  updateViewedAVErrorPage = (value: boolean) => {
+    this.hasViewedAVErrorPage = value
+  }
```
seniorquico, line 569: This looks like a bad code smell... the Salad Bowl store is managing state that it doesn't even use or care about. It would be better if this flag was maintained in the UI store, and the UI store would need to react to the start/end of a mining session.

packages/web-app/src/UIStore.ts
```diff
@@ -49,17 +50,28 @@ export class UIStore {
...
+        if (!hasViewedAVErrorPage && errorMessage) {
+          this.store.analytics.trackMiningError(errorMessage.errorCategory, errorMessage.errorCode)
+        }
+
         const antiVirusSoftware = this.store.zendesk.detectedAV
         if (antiVirusSoftware) {
           this.store.analytics.trackErrorPageViewed(`${antiVirusSoftware} Anti-Virus Error`)
           const articleId = getZendeskAVData(antiVirusSoftware).id
-          this.showModal(`/errors/anti-virus/${articleId}`)
+          if (!hasViewedAVErrorPage) {
```
seniorquico, line 65: First, it feels wasteful to send two events (`trackMiningError` and `trackErrorPageViewed`) at the same time. Can these not be condensed?

Second, why do we track the error page viewed when we _don't_ view the error page (i.e. `hasViewedAVErrorPage === true`)?

seniorquico (changes requested):

packages/web-app/src/UIStore.ts
scottcinatl: Just had a conversation with Eric about this and he is okay with having both events being fired for now.

seniorquico (approved):

PR #653, opened 2021-03-19
Update start button

scottcinatl:
1. Modifies time frames for Start Button prepping progress percentage changes
2. Updates Start Button styles to give a more animated effect while prepping

### DIFF
--- a/packages/web-app/src/modules/machine-views/components/StartButton.tsx
+++ b/packages/web-app/src/modules/machine-views/components/StartButton.tsx
@@ -1,70 +1,74 @@
-import classnames from 'classnames'
+import classNames from 'classnames'
 import { Component } from 'react'
 import withStyles, { WithStyles } from 'react-jss'
 import { SaladTheme } from '../../../SaladTheme'
 import { formatDuration } from '../../../utils'
 import { MiningStatus } from '../../machine/models'
 
+const buttonHeight = 40
 const buttonWidth = 120
-const buttonWidthLarge = 160
+const skew = 30
+const skewMarginleft = (Math.tan((skew * Math.PI) / 180) * buttonHeight) / 2
 
 const styles = (theme: SaladTheme) => ({
   startButtonContainer: {
     '-webkit-app-region': 'none',
     position: 'relative',
-    width: buttonWidth + 15,
-    paddingLeft: 10,
+    width: buttonWidth,
+    marginRight: 30,
     height: '2rem',
   },
-  startButton: {
-    position: 'absolute',
-    top: 0,
-    height: 40,
+  button: {
+    position: 'relative',
+    transform: `skewX(-${skew}deg)`,
     width: buttonWidth,
-    transform: 'skew(-30deg)',
-    background: '#B2D530',
+    height: buttonHeight,
+    marginLeft: skewMarginleft,
+    backgroundColor: '#B2D530',
     border: '1px solid rgba(201, 240, 55, 0.5)',
-    color: '#fff',
-    textShadow: '0px 0px 5px rgba(201, 240, 55, 0.9), -1px -1px 3px rgba(0, 0, 0, 0.25)',
-    backdropFilter: 'blur(4px)',
   },
-  startButtonInstalling: {
+  buttonPrepping: {
     background: '#DBF1C1',
   },
-  startButtonPrepping10: {
-    backgroundImage: 'linear-gradient(to right, #B2D530, #B2D530 10%, #DBF1C1 10%, #DBF1C1 90%)',
-  },
-  startButtonPrepping25: {
-    backgroundImage: 'linear-gradient(to right, #B2D530, #B2D530 25%, #DBF1C1 25%, #DBF1C1 75%)',
-  },
-  startButtonPrepping60: {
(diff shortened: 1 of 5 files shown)


packages/web-app/src/modules/machine-views/components/StartButton.tsx
```diff
@@ -175,47 +169,47 @@ class _StartButton extends Component<Props, State> {
...
-            [classes.startButtonLarge]: isLarge,
-            [classes.startButtonInstalling]: status === MiningStatus.Installing,
-            [classes.startButtonPrepping10]: status === MiningStatus.Initializing && progress <= 0.1,
-            [classes.startButtonPrepping25]: status === MiningStatus.Initializing && progress > 0.1 && progress <= 0.2,
-            [classes.startButtonPrepping60]: status === MiningStatus.Initializing && progress > 0.2 && progress <= 0.3,
-            [classes.startButtonPrepping90]: status === MiningStatus.Initializing && progress >= 0.4,
-            [classes.startButtonChopping]: status === MiningStatus.Running,
+          className={classNames(classes.button, {
+            [classes.buttonPrepping]: status === MiningStatus.Installing || status === MiningStatus.Initializing,
+            [classes.buttonChopping]: status === MiningStatus.Running,
```
seniorquico, line 182:
```ts
const isPrepping = status !== MiningStatus.Running && progress !== undefined
const label = isPrepping ? (status === MiningStatus.Installing ? MiningStatus.Installing : MiningStatus.Initializing) : status
```

packages/web-app/src/modules/salad-bowl/SaladBowlStore.ts
```diff
@@ -159,28 +161,45 @@ export class SaladBowlStore implements IPersistentStore {
...
 
     if (runningTime) {
-      if (runningTime <= 60000) {
-        progress = 0.1
+      if (runningTime <= 10000) {
+        percentage = runningTime / 1000
+      }
+
+      if (runningTime > 10000 && runningTime <= 30000) {
+        percentage += 0.25
```
seniorquico, line 173: Do not mutate the percentage. This should be derived/computed from the `runningTime`.

packages/web-app/src/modules/salad-bowl/SaladBowlStore.ts
```diff
@@ -56,6 +56,8 @@ export class SaladBowlStore implements IPersistentStore {
   @observable
   public hasViewedAVErrorPage: boolean = false
 
+  private preppingProgressPercentage: number = 0
```
seniorquico, line 59: Remove this!

seniorquico (changes requested):

packages/web-app/src/modules/salad-bowl/SaladBowlStore.ts
scottcinatl: I still need to store the state of the progress made during prepping rather than just reference the run time. Each second that passes during the prep time I need to update the delta in percentage being added to progress based on which time frame the run time falls under.  Having the stored progress state makes this easier to do. 

For example, between 10 seconds and 30 seconds of run time, we are jumping from 10% to 15%. During that time frame, we are applying a .25% change each second. Each time frame has different percentage applied per second.

packages/web-app/src/modules/machine-views/components/StartButton.tsx
```diff
@@ -21,8 +21,8 @@ const styles = (theme: SaladTheme) => ({
     position: 'relative',
     transform: `skewX(-${skew}deg)`,
     width: buttonWidth,
-    height: buttonHeight /* fixed */,
-    marginLeft: 11.547 /* fixed: tan(-30) * height / 2 */,
+    height: buttonHeight,
+    marginLeft: () => (Math.tan((skew * Math.PI) / 180) * buttonHeight) / 2,
```
seniorquico, line 25: The `skew` and `buttonHeight` are constants, and this should also be a constant value.

packages/web-app/src/modules/salad-bowl/SaladBowlStore.ts
seniorquico: Let's remove this! ;-)

seniorquico (changes requested):

packages/web-app/src/modules/machine-views/components/StartButton.tsx
```diff
@@ -1,70 +1,73 @@
...
 import { Component } from 'react'
 import withStyles, { WithStyles } from 'react-jss'
 import { SaladTheme } from '../../../SaladTheme'
 import { formatDuration } from '../../../utils'
 import { MiningStatus } from '../../machine/models'
 
+const buttonHeight = 40
 const buttonWidth = 120
-const buttonWidthLarge = 160
+const skew = 30
```
seniorquico, line 10:
```ts
const skewedMarginLeft = (Math.tan((skew * Math.PI) / 180) * buttonHeight) / 2;
```

seniorquico (approved):
It's beautiful!

seniorquico (approved):

PR #654, opened 2021-03-23
Added Android, Epic Games, and Nintendo platform icons

scottcinatl:


packages/web-app/src/modules/reward-views/components/RewardInfoPanel.tsx
```diff
@@ -5,7 +5,10 @@ import withStyles, { WithStyles } from 'react-jss'
 import { Divider } from '../../../components'
 import { SaladTheme } from '../../../SaladTheme'
 import { Reward, RewardPlatform } from '../../reward/models'
+import android from './assets/logos/android.png'
 import apple from './assets/logos/apple.png'
+import epicGames from './assets/logos/epicGames.png'
+import nintendo from './assets/logos/nintendo.png'
```
seniorquico, line 11: Do we have standards on size/aspect ratio? Seems odd that these are all different.

scottcinatl: I tried to put all three to 128 x 128, but the aspect ratio for epic games icons looks squished at that size, so that one is a little different.

seniorquico (changes requested):
Please drop in the optimized assets and add a quick readme/note to callout the needed dimensions.

seniorquico (approved):

PR #669, opened 2021-04-26
Remove Offerwall from App

jpark799:
~~**_This PR is not ready yet._**~~ 

~~Removed offerwall but still waiting on Arlo to provide revised images. 
`earning-overview.svg` and `how-salad-works-rewards.svg` need to be revised.~~

This is ready for review. All offerwall mentions have been pulled from the app and images have been updated.


packages/web-app/src/modules/error-views/components/AntiVirusFirewallErrorPage.tsx
```diff
@@ -347,7 +347,7 @@ class _AntiVirusFirewallErrorPage extends Component<Props, State> {
                             </p>
                           ) : (
                             <p>
-                              Your Anti-Virus is blocking you from chopping. We'll keep testing other miners, but your
+                              Your Firewall is blocking you from chopping. We'll keep testing other miners, but your
```
seniorquico, line 350: This isn't related to offerwalls. ;-)

It looks like this PRs going to get delayed based on business constraints... Should this change be pulled out into a separate PR?

packages/web-app/src/modules/storefront-views/utils.ts
```diff
@@ -301,14 +301,5 @@ export const strapiJSON = {
         },
       ],
     },
-    {
```
seniorquico, line 304: Is this data blob for Storybook? If so-- we should move it out of this file. Doesn't make sense to mix mock data in with other application utilities.

seniorquico (changes requested):

packages/web-app/src/modules/storefront-views/utils.ts
jpark799: yeah, that makes sense. i'll go ahead and pull this out of the PR. 

thanks

seniorquico (approved):

PR #683, opened 2021-05-17
Trimming whitespace in rewardValueItem.code

jpark799:
"Some codes have leading or trailing white space, need to remove this white space when we load the data for the reward vault"

### DIFF
--- a/packages/web-app/src/modules/vault/VaultStore.ts
+++ b/packages/web-app/src/modules/vault/VaultStore.ts
@@ -32,7 +32,7 @@ export class VaultStore {
     name: r.name,
     price: r.price,
     timestamp: new Date(r.timestamp),
-    code: r.code ? r.code.trim() : r.code,
+    code: r.code?.trim(),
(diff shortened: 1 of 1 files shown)


packages/web-app/src/modules/vault/VaultStore.ts
```diff
@@ -32,7 +32,7 @@ export class VaultStore {
     name: r.name,
     price: r.price,
     timestamp: new Date(r.timestamp),
-    code: r.code,
+    code: r.code ? r.code.trim() : r.code,
```
seniorquico, line 35: nit:

```suggestion
    code: r.code?.trim(),
```

seniorquico (approved):

packages/web-app/src/modules/vault/VaultStore.ts
jpark799: this feels much cleaner. thank you

seniorquico (approved):

PR #687, opened 2021-06-11
Incorporates Garden

jpark799:
Making this PR to get the Garden incorporated in our project so if Scott runs out of work to do, he isn't blocked and can just pull from operation-avalanche to get started on a page. 

It's important to note that we will need to redo this process each time we get new components published in the garden but we are all familiar with the process now lol.


packages/web-app/src/modules/reward-views/components/RewardDescriptionPanel.tsx
```diff
@@ -1,4 +1,4 @@
-import { DomElement } from 'htmlparser2'
+// import type { ElementType } from 'htmlparser2'
```
jpark799, line 1: I spent a lot of time digging into this because it was saying DomElement isn't exported. It turns out DomElement has been renamed to ElementType but when I tried to use it, I was getting a namespace error. Given our tight deadline and the fact that this is a thirdparty library, I made the decision to just use any and comment this out for now. please let me know if you would like to talk in more detail about this.

packages/web-app/src/modules/reward-views/components/RewardDescriptionPanel.tsx
```diff
@@ -1,4 +1,4 @@
-import { DomElement } from 'htmlparser2'
+// import { Element } from 'htmlparser2'
```
jpark799, line 1: I spent a lot of time digging into this because it was saying DomElement isn't exported. It turns out DomElement has been renamed to [Element ](https://github.com/fb55/htmlparser2/issues/601#issuecomment-707850808)(possibly even ElementType) but when I tried to use it, I was getting a namespace error. Given our tight deadline and the fact that this is a thirdparty library, I made the decision to just use any and comment this out for now. please let me know if you would like to talk in more detail about this.

seniorquico: Please remove the commented out code.

seniorquico (changes requested):

jpark799:
i realized we're exporting that entire themes file from the garden so i deleted the file from this folder so we only have to keep track of it in one place.

seniorquico (approved):

PR #689, opened 2021-06-15
Added basic bonus page

dsarfati:


packages/web-app/src/modules/bonus/BonusStore.ts
```diff
@@ -0,0 +1,48 @@
...
+      throw error
+    }
+  })
+
+  @action.bound
+  claimBonus = flow(function* (this: BonusStore, bonusId: string) {
+    console.log('Claiming bonus ' + bonusId)
+
+    try {
+      yield this.axios.post(`/api/v2/bonuses/${bonusId}/claim`)
```
seniorquico, line 40: On success, does it make sense to show a toast notification? The toast could even be cute wrt click handler... if it's a reward bonus, it could take you to the reward vault.

packages/web-app/src/index.tsx
```diff
@@ -1,28 +1,27 @@
...
 import { Router } from 'react-router-dom'
+import 'react-toastify/dist/ReactToastify.css'
+import 'url-polyfill'
+import 'whatwg-fetch'
 import { App } from './App'
 import { createClient } from './axiosFactory'
 import { Head } from './components'
 import { config } from './config'
 import { ErrorBoundary } from './ErrorBoundary'
+import './index.css'
```
seniorquico, line 24: This was sorted (probably something that VS Code did automatically). The order here is super important, and these should be put back to their original order.

seniorquico (changes requested):

packages/web-app/src/index.tsx
```diff
@@ -73,18 +72,29 @@ routerHistory.block((location, action) => {
...
-      <SkeletonTheme color={'#172E40'} highlightColor="#304759">
-        <ErrorBoundary>
-          {/* Default page title for any page that doesn't specify one */}
-          <Head title="Salad Technologies" />
-          <Tooltips />
-          <App history={history} />
-        </ErrorBoundary>
-      </SkeletonTheme>
-    </ThemeProvider>
+    <RawIntlProvider value={intl}>
```
scottcinatl, line 86: @dsarfati Had to add the RawIntlProvider on my PR as well. I did mine in App.tsx. So maybe now that mine is already merged in you could catch up with my branch and remove the ThemeProvider and RawIntlProvider from the App file

packages/web-app/src/modules/bonus/BonusStore.ts
seniorquico: Added a backlog task to review this.

seniorquico (approved):

PR #691, opened 2021-06-16
Adds Current Season Page

scottcinatl:
- This PR has now been tested against the API. A follow up PR will be needed to implement the correct Levels UI to the page

### DIFF
--- a/packages/web-app/src/modules/seasons-views/pages/CurrentSeasonPage.tsx
+++ b/packages/web-app/src/modules/seasons-views/pages/CurrentSeasonPage.tsx
@@ -10,7 +10,9 @@ const styles = (theme: SaladTheme) => ({
     color: theme.darkBlue,
     display: 'flex',
     flexDirection: 'column',
-    height: '100vh',
+  },
+  levels: {
+    display: 'flex',
   },
   subtitle: {
     alignItems: 'center',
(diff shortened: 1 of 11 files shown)


packages/web-app/src/modules/seasons-views/pages/CurrentSeasonPage.tsx
```diff
@@ -0,0 +1,74 @@
...
+import { SaladTheme } from '../../../SaladTheme'
+import { withLogin } from '../../auth-views'
+import { Level } from '../../seasons/models'
+
+const styles = (theme: SaladTheme) => ({
+  container: {
+    color: theme.darkBlue,
+    display: 'flex',
+    flexDirection: 'column',
+    height: '100vh',
```
seniorquico, line 13: This makes the page bigger than it should be (forces scroll bars) if the content is shorter than the full height of the viewport because it does not account for the extra space taken up by the nav bar or the window controls.

See Jason's `Account.tsx` file (from the profile PR) & Daniel's `BonusPage.tsx` file (from the bonus PR). The "page" element needs the following styles:

```css
flex: 1,
backgroundImage: 'linear-gradient(to right, #56A431 , #AACF40)'
```

seniorquico (changes requested):

seniorquico (approved):

PR #694, opened 2021-06-16
Added notifications when bonuses are claimed

dsarfati:


packages/web-app/src/modules/bonus/BonusStore.ts
```diff
@@ -38,8 +39,48 @@ export class BonusStore {
 
     try {
       yield this.axios.post(`/api/v2/bonuses/${bonusId}/claim`)
+
+      const bonus = this.unclaimedBonuses.find((x) => x.id === bonusId)
```
seniorquico, line 43: There's a race condition here. Because the HTTP request to claim it is async, you cannot guarantee that this is the next line of code to execute. While this request is happening, the refresh service's timer could've already been queued to fire and update our list of unclaimed bonuses. It's possible this could return `undefined`.

You should pull this line to the top of the function, before the async request.

seniorquico (changes requested):

seniorquico (approved):

PR #719, opened 2021-06-21
Allow default salad avatar selection

jpark799:
I remember hearing we don't want to add the default avatar to the api call because it could end up getting expensive when we could just add it on client side. This was a way I found to do it but if you have any suggestions to make this cleaner, please feel free to let me know :)


dsarfati (approved):
Looking good!

PR #728, opened 2021-07-06
Added logic that allows user to decide whether or not to replace current earning rate bonus

scottcinatl:


packages/web-app/src/modules/bonus/BonusStore.ts
```diff
@@ -94,8 +104,13 @@ export class BonusStore {
 
       // Check to see if we are overriding a current earning rate bonus
       if (bonus.blockType === BonusType.EarningRate && this.currentEarningBonus !== undefined) {
-        // TODO: Scott. Show a modal here warning the user that they are going to replace {this.currentEarningBonus.multiplier}x with a new bonus {bonus.multiplier}x
-        console.error('TODO: Need confirmation that the user wants to replace their earning rate')
+        try {
+          yield this.replaceCurrentEarningRatePrompt()
```
seniorquico, line 108: The promise doesn't get canceled if the component is unmounted prematurely (say another component kicks off a route change... the auth store "signs out" the user... Salad Bowl store pops open the AV or incompatible modals... etc.).

I'm not seeing the value of the promise here. It's an internal implementation detail of the store, and it appears to be complicating the flow. With a slight refactor (and no changes to the components at all!), this can be accomplished without the promise.

I think we should remove this, and wait to introduce it when we have a legit, shared dialog manager component. In that scenario, I see the promise being a cross-store communication method.

seniorquico (changes requested):

packages/web-app/src/modules/bonus/BonusStore.ts
scottcinatl: @seniorquico Removed the promise-based replacement flow

seniorquico (approved):

PR #735, opened 2021-07-12
Carousel Scroll for Levels

jpark799:
[React-Multi-Carousel unfortunately does not allow you to start at a non-zero index](https://github.com/YIZHUANG/react-multi-carousel/issues/150)  so I had to reorder the array to have the current level be the first item. I had to make the carousel infinite so if you go to the left of your current level, you can still see the earlier levels.

### DIFF
--- a/packages/web-app/src/modules/seasons-views/pages/CurrentSeasonPage.tsx
+++ b/packages/web-app/src/modules/seasons-views/pages/CurrentSeasonPage.tsx
@@ -3,6 +3,7 @@ import { useMemo } from 'react'
 import Scrollbars from 'react-custom-scrollbars'
 import { useIntl } from 'react-intl'
 import withStyles, { WithStyles } from 'react-jss'
+import Carousel from 'react-multi-carousel'
 import { Head } from '../../../components'
 import { SaladTheme } from '../../../SaladTheme'
 import { withLogin } from '../../auth-views'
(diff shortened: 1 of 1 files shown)


packages/web-app/src/modules/seasons-views/pages/CurrentSeasonPage.tsx
```diff
@@ -94,6 +99,9 @@ const _CurrentSeasonPage = ({
         }))
       : []
   }, [levels, nextLevel, currentLevelXP])
+
+  nextLevel >= 2 && levelCards.push(...levelCards.splice(0, nextLevel - 1))
```
dsarfati, line 103: What does this code do?? Is this moving the cards around?

dsarfati (changes requested):
Are those the default arrows? Is there an issue using the same arrows we use for the reward list?

packages/web-app/src/modules/seasons-views/pages/CurrentSeasonPage.tsx
jpark799: yeah, it's moving the current level to the front of the array and moving everything before the current array to the back of the array

jpark799:
image

dsarfati (approved):

PR #743, opened 2021-07-26
Update Override Compatibility Flow

scottcinatl:
- Removes disabled checkbox state from checkbox to override Salad's compatibility detection
- Shows modal warning user about enabling the override compatibility detection feature
- Adds overridden property to Mixpanel start event


packages/web-app/src/modules/analytics/AnalyticsStore.ts
```diff
@@ -21,6 +21,7 @@ export class AnalyticsStore {
 
     const token = config.mixpanelToken
     if (!token) {
+      mixpanel.init('TEST TOKEN')
```
scottcinatl, line 24: @seniorquico @jpark799 Not sure if this is the best solution for resolving the current Mixpanel issues while in local development. But this change does at least help initializing the Mixpanel configuration locally and eliminates all the client side app crashes.

seniorquico (approved):

PR #744, opened 2021-07-28
Onboarding Phase 1

jpark799:
First phase of Onboarding update

- Updated Referral page designs
- Using garden checkbox on Login page
- Removed navbar from Login page
- Login page is automatically opened if a user is not signed in when they open the app in the desktop app
- mixpanel event tracking added
  - tracking each page view in the onboarding process with page name and order number (onboarding process starts on welcome page).
  - tracking `Let's go` button clicks on Welcome page and `Give me a bonus!` clicks on Referral Code page.
  - tracking promo code enters along with any error message that may follow on Referral Code page. 
- Implemented loading screen that shows for at least 2 seconds (so it doesn't flash and confuse users).


packages/web-app/src/modules/referral/ReferralStore.ts
```diff
@@ -123,15 +141,16 @@ export class ReferralStore {
...
+  submitDefaultReferralCodeOnboarding = flow(function* (this: ReferralStore) {
+    this.store.analytics.trackButtonClicked('give_me_a_bonus_button', `Give me a bonus! button`, 'enabled')
     yield this.submitReferralCode('SALAD')
   })
 
-  /** Called when a user enters in a referral code */
+  /** Called when a user enters in a referral code on Onboarding Page */
   @action.bound
-  submitReferralCode = flow(function* (this: ReferralStore, code: string) {
+  submitReferralCodeOnboarding = flow(function* (this: ReferralStore, code: string) {
```
jpark799, line 153: `submitReferralCode` is being used in the account page so in order to track only the referral codes being submitted in the onboarding process, I made a new function called `submitReferralCodeOnboarding` and renamed `submitDefaultReferralCode` to submitDefaultReferralCodeOnboarding` to be more specific

packages/web-app/src/index.tsx
```diff
@@ -86,6 +86,10 @@ const intl = createIntl(
   cache,
 )
 
+if (!rootStore.auth.isAuthenticated) {
```
seniorquico, line 89: Upfront login is required when running in Electron. However, login may be deferred when running in a web browser.

packages/web-app/src/modules/analytics/AnalyticsStore.ts
```diff
@@ -328,7 +328,7 @@ export class AnalyticsStore {
     })
   }
 
-  /** Track when a button is clicked */
+  /** Track when a button is clicked */ // use this for button clicks
```
scottcinatl, line 331: You could remove this added comment.

packages/web-app/src/modules/analytics/AnalyticsStore.ts
```diff
@@ -392,6 +392,26 @@ export class AnalyticsStore {
...
+  }
+
+  /** Track when a referral code is entered in the onboarding flow */
+  public trackReferralCodeEntered = (code: string, isSuccess: boolean, errorMessage?: string) => {
+    this.track('Referral Code Entered On User Attribution Page', {
+      referralCode: code,
+      isSuccess: isSuccess,
+      errorMessage: errorMessage,
+    })
+  }
```
seniorquico, line 413: Remove this. The backend already pushes this event.

seniorquico (changes requested):

packages/web-app/src/modules/referral/ReferralStore.ts
```diff
@@ -111,7 +123,8 @@ export class ReferralStore {
 
       //If we haven't entered a referral code, show the onboarding page
       if (!this.currentReferral) {
-        this.store.ui.showModal('/onboarding/referral')
+        this.store.ui.showModal('/onboarding/welcome')
+        this.store.analytics.trackOnboardingPageViewed('Onboarding Welcome', 1) // pass username through?
```
scottcinatl, line 127: You could also remove the comment here

packages/web-app/src/index.tsx
jpark799: oh that's a great point. thank you for catching that

jpark799: i'm doing a `if (window.salad.platform !== 'web' && !rootStore.auth.isAuthenticated)` to make sure we're not in a web browser. if there's another preferred way to do this check, please let me know but i just tested and it's working as we need.

scottcinatl: Instead of `.window.salad.platform !== 'web'`, we already have a check to see if the user is on the desktop app with `this.store.native.isNative`

jpark799: @scottcinatl oh awesome. thank you for letting me know.

packages/web-app/src/index.tsx
```diff
@@ -86,6 +86,10 @@ const intl = createIntl(
   cache,
 )
 
+if (rootStore.native.isNative && !rootStore.auth.isAuthenticated) {
```
seniorquico, line 89: If I recall correctly, `isAuthenticated` is determined _asynchronously_ (it may require an API call to refresh the access token). How does routing to `/login` work after the `isAuthenticated` flag is set to `true`?

jpark799: I'm just doing a `rootStore.routing.replace('/login')` to set the route to the login page. I did some digging through the codebase and saw this is how we're doing it in a number of places. If there's another way to do routing that you would prefer, please let me know and I'd be happy to make an update.

packages/web-app/src/Store.ts
```diff
@@ -43,6 +43,9 @@ export const createStore = (axios: AxiosInstance): RootStore => {
 export const getStore = (): RootStore => sharedStore
 
 export class RootStore {
+  @observable
+  public appLoaded: boolean = true
```
seniorquico, line 47: Rename to `appLoading`.

packages/web-app/src/index.tsx
```diff
@@ -12,9 +12,10 @@ import 'url-polyfill'
 
 // Import dependencies.
 import { ThemeProvider as EmotionThemeProvider } from '@emotion/react'
-import { DefaultTheme as EmotionTheme } from '@saladtechnologies/garden-components'
+import { DefaultTheme as EmotionTheme, LoadingScreen } from '@saladtechnologies/garden-components'
```
seniorquico, line 15: The build failed because the new garden package was updated correctly:

- The old `tgz` file should be deleted.
- The `package.json` file should be updated to reference the new version.
- The `yarn.lock` file should be updated to reference the new version.

seniorquico (changes requested):

seniorquico (dismissed):

seniorquico (approved):

PR #754, opened 2021-08-09
Autorun fix

jpark799:
our `isAuthenticated` flag was previously set to an initial state of false even if this wasn't the actual state and this caused our `autoRun` function to be run twice (which would make `onLogout` run). Once on initial run, and once again after the `isAuthenticated` flag has been accurately set. This PR sets `isAuthenticated` to an initial state of undefined and exits the `autoRun` function if undefined which ensures we run `onLogin` or `onLogout` accurately.

### DIFF
--- a/packages/web-app/src/modules/salad-bowl/SaladBowlStore.ts
+++ b/packages/web-app/src/modules/salad-bowl/SaladBowlStore.ts
@@ -110,7 +110,7 @@ export class SaladBowlStore implements IPersistentStore {
   }
 
   @computed
-  get canRun(): boolean {
+  get canRun(): boolean | undefined {
     return (
       this.store.auth.isAuthenticated &&
       this.store.machine &&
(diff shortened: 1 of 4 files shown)


packages/web-app/src/modules/salad-bowl/SaladBowlStore.ts
```diff
@@ -110,7 +110,7 @@ export class SaladBowlStore implements IPersistentStore {
   }
 
   @computed
-  get canRun(): boolean {
+  get canRun(): boolean | undefined {
     return (
       this.store.auth.isAuthenticated &&
```
seniorquico, line 115:
```suggestion
  get canRun(): boolean {
    return (
      this.store.auth.isAuthenticated !== undefined &&
      this.store.auth.isAuthenticated &&
```

seniorquico (changes requested):

seniorquico (approved):

PR #759, opened 2021-08-13
Disable Sleep Mode

jpark799:


packages/desktop-app/src/main.ts
```diff
@@ -313,7 +313,59 @@ const createMainWindow = () => {
 
   ipcMain.on('js-dispatch', bridge.receiveMessage)
 
-  //Listen for machine info requests
+  bridge.on('disable-sleep-mode', () => {
+    async function spawnChild() {
+      const { spawn } = require('child_process')
```
seniorquico, line 318: Imports should be up at the top of the file (unless there's a specific reason to delay the import, but there shouldn't be one for this use case).

packages/desktop-app/src/main.ts
```diff
@@ -313,7 +313,59 @@ const createMainWindow = () => {
...
+      // Initial command that prints off GUIDS of all power profiles
+      const child = spawn('powercfg /list', {
+        windowsHide: true,
+      })
+
+      let data = ''
+      for await (const chunk of child.stdout) {
```
seniorquico, line 325: `child_process.exec` can do this for you, and the output will be passed to a callback you define.

packages/desktop-app/src/main.ts
```diff
@@ -313,7 +313,59 @@ const createMainWindow = () => {
...
+
+      if (exitCode) {
+        throw new Error(`subprocess error exit ${exitCode}, ${error}`)
+      }
+      return data
+    }
+
+    spawnChild().then(
+      (data) => {
+        const regex = /\:\s(.*?)\(/g
```
seniorquico, line 346: This pattern should probably be a little more forgiving on whitespace and a little more strict on the GUID matching.

packages/desktop-app/src/main.ts
```diff
@@ -313,7 +313,59 @@ const createMainWindow = () => {
...
+        const GUIDS = data.match(regex)
+        // loops through all the GUIDS and executes process to disable sleepmode
+        GUIDS?.forEach((element) => {
+          const GUID = element.substring(2, element.length - 3)
+          const { exec } = require('child_process')
```
seniorquico, line 351: Imports should be up at the top of the file (unless there's a specific reason to delay the import, but there shouldn't be one for this use case).

packages/desktop-app/src/main.ts
```diff
@@ -313,7 +313,59 @@ const createMainWindow = () => {
...
+      return data
+    }
+
+    spawnChild().then(
+      (data) => {
+        const regex = /\:\s(.*?)\(/g
+        const GUIDS = data.match(regex)
+        // loops through all the GUIDS and executes process to disable sleepmode
+        GUIDS?.forEach((element) => {
+          const GUID = element.substring(2, element.length - 3)
```
seniorquico, line 350: Use capture groups so the data is returned exactly as needed.

packages/desktop-app/src/main.ts
```diff
@@ -313,7 +313,59 @@ const createMainWindow = () => {
...
+          const { exec } = require('child_process')
+          const disableSleepCommand = exec(
```
seniorquico, line 352: We may want to run these different calls sequentially. The powercfg docs do not specify how it handles concurrent calls. It may work fine, but this could also be a source of subtle errors if it does not.

packages/desktop-app/src/main.ts
```diff
@@ -313,7 +313,59 @@ const createMainWindow = () => {
...
+            `powercfg /setacvalueindex ${GUID} 238c9fa8-0aad-41ed-83f4-97be242c8f20 29f6c1db-86da-48c5-9fdb-f2b67b1f44da 0`,
+            {
+              windowsHide: true,
+            },
+          )
+          disableSleepCommand.on('exit', function () {
+            console.log('Child process exited with exit code ')
```
seniorquico, line 359: Please remove log statements that don't help our support team diagnose a problem. For example... this is an expected situation (the process will always exit), and it doesn't tell the support team if it succeeded or failed. As is, this just fills the log file with additional data that must be reviewed and ignored.

seniorquico (changes requested):

PR #765, opened 2021-08-18
Add Disable Sleep Mode Function

jpark799:
This PR adds the ability to enable sleep mode on all the user's power profiles. We will still need to wire this up to the frontend.

### DIFF
--- a/packages/desktop-app/src/main.ts
+++ b/packages/desktop-app/src/main.ts
@@ -1,4 +1,5 @@
 import * as Sentry from '@sentry/electron'
+import { exec } from 'child_process'
 import {
   app,
   BrowserWindow,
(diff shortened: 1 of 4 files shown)


packages/desktop-app/src/main.ts
```diff
@@ -314,6 +315,64 @@ const createMainWindow = () => {
   ipcMain.on('js-dispatch', bridge.receiveMessage)
 
   //Listen for machine info requests
+  bridge.on('disable-sleep-mode', () => {
```
seniorquico, line 318: The comment above this doesn't look right.

packages/desktop-app/src/main.ts
```diff
@@ -314,6 +315,64 @@ const createMainWindow = () => {
...
+    const powercfg = 'C:\\Windows\\System32\\powercfg.exe'
```
seniorquico, line 319: We should add a check to guard against running this on macOS/Linux platforms.

packages/web-app/src/modules/machine/NativeStore.ts
```diff
@@ -141,6 +142,24 @@ export class NativeStore {
     window.salad.dispatch(type, payload && toJS(payload))
   }
 
+  disableSleepMode = (): Promise<void> => {
```
seniorquico, line 145: We should add a check to guard against running this on macOS/Linux platforms.

packages/web-app/src/modules/machine/NativeStore.ts
```diff
@@ -141,6 +142,24 @@ export class NativeStore {
...
+          } else {
+            reject()
+          }
+        })
+        this.send(disableSleepMode)
+      })
+    } else {
+      return Promise.reject('The process is already running.')
+    }
+  }
```
seniorquico, line 161: We may want to consider adding a `canDisableSleepMode` property. That would be computed using the current platform & the version of the native bridge interface. The onboarding business logic would use this to know if the sleep mode page should be shown or automatically skipped when unsupported.

seniorquico (changes requested):

packages/web-app/src/modules/machine/NativeStore.ts
jpark799: by version of the native bridge interface, do you mean to just make sure the desktop version is greater than 13.1.2(current version) or would the api version also work(currently 8)?

packages/web-app/src/modules/machine/NativeStore.ts
```diff
@@ -141,6 +145,28 @@ export class NativeStore {
...
+  disableSleepMode = (): Promise<void> => {
+    if (this.canDisableSleepMode) {
+      if (!this.callbacks.has(disableSleepMode)) {
+        return new Promise((resolve, reject) => {
+          this.callbacks.set(disableSleepMode, (result: { success: boolean }) => {
+            this.callbacks.delete(disableSleepMode)
+            if (result.success) {
+              resolve()
+            } else {
+              reject()
```
seniorquico, line 157: The `reject`s below return a message. Should this also return a generic error message? Something to the effect of "try again" and "if it keeps failing, contact support"?

packages/desktop-app/src/main.ts
```diff
@@ -313,6 +314,64 @@ const createMainWindow = () => {
...
+      },
+      (error, stdout, stderr) => {
+        if (error == null) {
+          const profiles = [...stdout.matchAll(/\:\s+([^\s]+)/g)]
+          disableSleepMode(profiles.shift())
+
+          function disableSleepMode(profile: RegExpMatchArray | undefined) {
+            if (profile) {
+              exec(
+                `${powercfg} /setacvalueindex ${profile[1]} 238c9fa8-0aad-41ed-83f4-97be242c8f20 29f6c1db-86da-48c5-9fdb-f2b67b1f44da 0`,
```
seniorquico, line 334: Add comment(s) to describe what each of these magic GUIDs stands for.

seniorquico (changes requested):

seniorquico (approved):

PR #766, opened 2021-08-19
Add Whitelist Windows Defender Function 

jpark799:
This PR adds the ability to whitelist windows defender. We will still need to wire this up to the frontend but the function calls and script is all there now.


packages/desktop-app/src/main.ts
```diff
@@ -314,6 +316,41 @@ const createMainWindow = () => {
...
+        `"'`,
+      {
+        env: {},
+        timeout: 60000,
+        windowsHide: true,
+      },
+      (error, _stdout, _stderr) => {
+        if (error) {
+          console.error(`Failed to Whitelist Windows Defender: ${error}`)
+          const phrase = /The operation was canceled by the user./
```
jpark799, line 335: We print off `The operation was canceled by the user` in the error message when users select no after being prompted by the elevated command. I'm just using regex to see if this phrase is in the error message and if not, we can safely assume that there was caused by the script.

packages/desktop-app/src/models/WhitelistWindowsDefenderErrorType.ts
```diff
@@ -0,0 +1,4 @@
+export enum WHITELIST_WINDOWS_DEFENDER_ERROR_TYPE {
```
jpark799, line 1: i saw we were doing screaming snake case in some places and just pascal case with naming our enums. i went with screaming snake case but if there's another pattern you would prefer, i can update this

packages/desktop-app/src/models/WhitelistWindowsDefenderErrorType.ts
```diff
@@ -0,0 +1,4 @@
...
+  USER_SELECTED_NO = 'User Selected No',
+  GENERAL_SCRIPT_ERROR = 'General Script Error',
```
jpark799, line 3: not super crazy about these error type names. if there's something that you think would fit better, i'm all ears.

packages/desktop-app/src/main.ts
seniorquico: The problem with the RegExp is that this message will be localized when Windows is set to use a different language. We should use the `error.code` property instead.

Testing & observing return values from various use cases will be the quickest way to identify what we're looking for. Microsoft documents all of the system error codes here:

https://docs.microsoft.com/en-us/windows/win32/debug/system-error-codes

jpark799: yeah, i thought about this too after our conversation around localized machines. unfortunately, clicking no on the prompt registers as an error code 1 which just means incorrect function. it doesn't look like there's a specific error code for when a user clicks no. running an invalid script, also prints off a code 1 so we wouldn't have a way to differentiate between a user clicking no and just a general error. the only difference i could find was the errormessage actually prints off ` This command cannot be run due to the error: The operation was canceled by the user.`

packages/desktop-app/src/main.ts
```diff
@@ -324,6 +326,37 @@ const createMainWindow = () => {
...
+          WHITELIST_DIR: nonDefaultFilePath
+            ? `${nonDefaultFilePath}\\Salad\\plugin-bin`
+            : `${process.env.APPDATA}\\Salad\\plugin-bin`,
+        },
+        timeout: 60000,
+        windowsHide: true,
+      },
+      (error, _stdout, _stderr) => {
+        if (error) {
+          console.error(`Failed to Whitelist Windows Defender: ${error.message}`)
```
seniorquico, line 343:
```suggestion
      (error, _stdout, stderr) => {
        if (error) {
          console.error(`Failed to Whitelist Windows Defender (${error.message}${error.code == null ? '' : `, exit code ${error.code}`})${stderr ? `\n${stderr}` : ''}`)
```

packages/desktop-app/src/main.ts
```diff
@@ -324,6 +326,37 @@ const createMainWindow = () => {
...
+      {
+        env: {
+          WHITELIST_DIR: nonDefaultFilePath
+            ? `${nonDefaultFilePath}\\Salad\\plugin-bin`
+            : `${process.env.APPDATA}\\Salad\\plugin-bin`,
```
seniorquico, line 336: If it isn't already, import the `path` module and use the `join` function.

```suggestion
            ? path.join(nonDefaultFilePath, 'Salad/plugin-bin')
            : path.join(process.env.APPDATA, 'Salad/plugin-bin'),
```

seniorquico (changes requested):

packages/desktop-app/src/main.ts
```diff
@@ -324,6 +326,37 @@ const createMainWindow = () => {
...
...oAMABTAFcANQBtAGIAeQBBADkASQBDAFIAcABEAFEAcAAwAGMAbgBrAGcAZQB3ADAASwBJAEMAQQBrAGMAQwA1AFQAZABHAEYAeQBkAEMAZwBwAEQAUQBwADkASQBHAE4AaABkAEcATgBvAEkASABzAE4AQwBpAEEAZwBKAEcAVQBnAFAAUwBBAGsAUgBYAEoAeQBiADMASgBiAE0ARgAwAE4AQwBpAEEAZwBhAFcAWQBnAEsAQwBSAGwASQBDADEAdQBaAFMAQQBrAGIAbgBWAHMAYgBDAEEAdABZAFcANQBrAEkAQwBSAGwATABrAFYANABZADIAVgB3AGQARwBsAHYAYgBpAEEAdABiAG0AVQBnAEoARwA1ADEAYgBHAHcAZwBMAFcARgB1AFoAQwBBAGsAWgBTADUARgBlAEcATgBsAGMASABSAHAAYgAyADQAdQBTAFcANQB1AFoAWABKAEYAZQBHAE4AbABjAEgAUgBwAGIAMgA0AGcATABXADUAbABJAEMAUgB1AGQAVwB4AHMASQBDADEAaABiAG0AUQBnAEoARwBVAHUAUgBYAGgAagBaAFgAQgAwAGEAVwA5AHUATABrAGwAdQBiAG0AVgB5AFIAWABoAGoAWgBYAEIAMABhAFcAOQB1AEwAawA1AGgAZABHAGwAMgBaAFUAVgB5AGMAbQA5AHkAUQAyADkAawBaAFMAQQB0AGIAbQBVAGcASgBHADUAMQBiAEcAdwBwAEkASABzAE4AQwBpAEEAZwBJAEMAQgBsAGUARwBsADAASQBDAFIAbABMAGsAVgA0AFkAMgBWAHcAZABHAGwAdgBiAGkANQBKAGIAbQA1AGwAYwBrAFYANABZADIAVgB3AGQARwBsAHYAYgBpADUATwBZAFgAUgBwAGQAbQBWAEYAYwBuAEoAdgBjAGsATgB2AFoARwBVAE4AQwBpAEEAZwBmAFMAQgBsAGIASABOAGwASQBIAHMATgBDAGkAQQBnAEkAQwBCAGwAZQBHAGwAMABJAEQARQBOAEMAaQBBAGcAZgBRADAASwBmAFEAMABLACIAfQAnACAAfAAgAEMAbwBuAHYAZQByAHQARgByAG8AbQAtAEoAcwBvAG4AKQAuAFMAYwByAGkAcAB0ACkAKQAgAHwAIABpAGUAeAA= -ExecutionPolicy Bypass',
```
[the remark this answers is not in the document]
jpark799: is there a reason you want it to be `index.js` and not `index.ts`?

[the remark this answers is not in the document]
jpark799: putting the index file as a `.ts` because i was getting this error. 

```  ERROR in undefined(undefined,undefined):
  TS5055: Cannot write file 'C:/Users/jason/Documents/Coding Applications/salad-applications/packages/desktop-app/src/whitelistWindowsScript/index.js' because it would overwrite input file.```

File: packages/desktop-app/src/main.ts (hunk 5 above)
Comment (jpark799) [reply packages/desktop-app/src/main.ts:396]: `process.env.APPDATA` is possibly undefined so we aren't able to use the `path.join`. i can have this default to something else if proces.env.APPDATA is undefined but what should i default it to?

Comment (seniorquico) [reply packages/desktop-app/src/main.ts:396]: It's technically possible to have an undefined APPDATA environment variable, although incredibly unlikely. You could handle it... if it's not defined, you can push an error to the web side.

File: packages/desktop-app/src/main.ts [hunk 7]
```diff
@@ -385,6 +386,40 @@ const createMainWindow = () => {
     })
   })
 
+  bridge.on('whitelist-windows-defender', (nonDefaultFilePath?: string) => {
+    exec(
+      powershellWhitelistCommand,
+      {
+        env: {
+          WHITELIST_DIR: nonDefaultFilePath
+            ? path.join(nonDefaultFilePath, 'Salad/plugin-bin')
```
seniorquico, line 395:
```suggestion
            ? path.normalize(filePath)
```

We should take the full file path from the caller. This will give us more flexibility, and can be used in the very near future with Salad Bowl's new folder paths (which won't have `plugin-bin` in the path).

seniorquico (changes requested):

seniorquico (approved):

PR #767, opened 2021-08-20
Added Salad Bowl 2.0 Login and Logout

scottcinatl:
- Adds `gprc-web` files to project to interact with Salad Bowl 2.0
- Adds initial structure of for Salad Bowl Store 2.0 along with the ability to login and logout


packages/web-app/src/services/SaladFork/SaladFork.tsx
```diff
@@ -0,0 +1,57 @@
...
+  authenticationPending: boolean
+  authenticationError: boolean
+}
+
+export class SaladFork implements SaladForkInterface {
+  readonly client: SaladBowlServices.SaladBowlServicePromiseClient
+
+  public isAuthenticated = false
+  public authenticationPending = false
+  public authenticationError = false
```
seniorquico, line 17: The fields should be private, or the state should be moved to a MobX store. If this is supposed to be a service client, the methods should probably return promises or, for more advanced scenarios, results communicated through RxJS observables or callbacks/listeners.

seniorquico (changes requested):

PR #772, opened 2021-08-27
Adds Sleep-Mode and Auto-Start pages to onboarding flow 

jpark799:


scottcinatl (approved):

packages/desktop-app/src/preload.js
```diff
@@ -21,7 +21,7 @@ function init() {
 
   window.salad = {
     platform: process.platform,
-    apiVersion: 9,
+    apiVersion: 8,
```
seniorquico, line 24: Why is this being rolled back?

seniorquico (changes requested):

packages/desktop-app/src/preload.js
jpark799: because we didn't have the latest desktop app published yet and we were using this as a flag to know if we should show the sleepmode pages. given that we've published the latest version of desktop, this can be 9 now.

scottcinatl (approved):

PR #773, opened 2021-08-30
Initial start/stop and data stream of salad bowl 2.0

scottcinatl:


packages/web-app/package.json
```diff
@@ -147,6 +148,7 @@
     "react-toastify": "7.0.4",
     "react-transition-group": "4.4.2",
     "recharts": "2.1.0",
+    "rxjs": "^7.3.0",
```
seniorquico, line 151:
```suggestion
    "rxjs": "7.3.0",
```

packages/web-app/src/modules/salad-bowl/SaladForkAndBowlStore.tsx
```diff
@@ -71,54 +95,313 @@ export class SaladForkAndBowlStore implements SaladBowlStoreInterface {
...
+  @action private resetPluginInfo = () => {
+    this.plugin = new PluginInfo()
   }
 
-  onDataLoaded = (data: unknown) => {
-    console.log(data)
+  private streamWorkloadData = () => {
+    this.store.saladFork
+      .workloadStatuses$()
+      .pipe(map((js) => js.getEvent()!))
```
seniorquico, line 140: It's possible `getEvent` returns `undefined`. We should not be using `!`.

seniorquico (changes requested):

seniorquico (approved):

PR #787, opened 2021-09-03
Antivirus 2 second wait

jpark799:
We're now waiting 2 seconds after successfully whitelisting windows defender to show the success toast notification and push users to the store. Also updated one of the error messages

### DIFF
--- a/packages/web-app/src/modules/onboarding/OnboardingAntivirusStore.tsx
+++ b/packages/web-app/src/modules/onboarding/OnboardingAntivirusStore.tsx
@@ -1,6 +1,6 @@
 import { action, flow, observable } from 'mobx'
 import { RootStore } from '../../Store'
-import { routeLink } from '../../utils'
+import { delay, routeLink } from '../../utils'
 import { NotificationMessageCategory } from '../notifications/models'
 import type { ZendeskArticle, ZendeskArticleList, ZendeskArticleResource } from '../zendesk/models'
 import { AntiVirusSoftware } from '../zendesk/models'
@@ -94,6 +94,7 @@ export class OnboardingAntivirusStore {
     this.whitelistWindowsDefenderPending = true
     try {
       yield this.store.native.whitelistWindowsDefender()
+      yield delay(2000)
       this.store.notifications.sendNotification({
         category: NotificationMessageCategory.Success,
         title: 'You’ve successfully whitelisted Salad!',
--- a/packages/web-app/src/modules/onboarding/OnboardingAntivirusStore.tsx
+++ b/packages/web-app/src/modules/onboarding/OnboardingAntivirusStore.tsx
@@ -94,16 +94,14 @@ export class OnboardingAntivirusStore {
     this.whitelistWindowsDefenderPending = true
     try {
       yield this.store.native.whitelistWindowsDefender()
-      setTimeout(() => {
-        this.store.notifications.sendNotification({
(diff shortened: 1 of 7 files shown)


packages/web-app/src/modules/onboarding/OnboardingAntivirusStore.tsx
```diff
@@ -94,14 +94,16 @@ export class OnboardingAntivirusStore {
...
       yield this.store.native.whitelistWindowsDefender()
-      this.store.notifications.sendNotification({
-        category: NotificationMessageCategory.Success,
-        title: 'You’ve successfully whitelisted Salad!',
-        message:
-          'Press the Start button to begin earning. The initial setup will then happen behind the scenes. This can take up to 30 minutes to complete.',
-        autoClose: 5000,
-      })
-      this.store.onboarding.viewNextPage(ONBOARDING_PAGE_NAMES.ANTIVIRUS_CONFIGURATION)
+      setTimeout(() => {
```
seniorquico, line 97: This breaks out of the flow. The delay should instead be part of it.

packages/web-app/src/modules/onboarding/models/WhitelistWindowsDefenderErrorType.ts
```diff
@@ -1,4 +1,4 @@
 export enum WhitelistWindowsDefenderErrorType {
-  USER_SELECTED_NO = `You’ll need to click ‘Yes’ on the Microsoft Access Control screen to whitelist. Please try again.`,
+  USER_SELECTED_NO = `You need to click 'Yes' on the User Account Control Prompt to whitelist Windows Defender. Please try again.`,
```
seniorquico, line 2: It seems weird that our long-form messages are our enum values.

seniorquico (changes requested):

packages/web-app/src/modules/onboarding/models/WhitelistWindowsDefenderErrorTypeMessage.ts
```diff
@@ -0,0 +1,4 @@
+export enum WhitelistWindowsDefenderErrorTypeMessage {
+  USER_SELECTED_NO = `You need to click 'Yes' on the User Account Control Prompt to whitelist Windows Defender. Please try again.`,
```
Vukkyy, line 2: whitelist... Windows Defender? I feel like this kind of implies you're whitelisting the Windows Defender application instead of whitelisting Salad (from what I've seen, it's referred to as "whitelisting Salad" everywhere else)

packages/web-app/src/modules/onboarding/models/WhitelistWindowsDefenderErrorType.ts
jpark799: i renamed WhitelistWindowsDefenderErrorType to WhitelistWindowsDefenderErrorTypeMessage to be more accurate.

packages/web-app/src/modules/onboarding/models/WhitelistWindowsDefenderErrorTypeMessage.ts
jpark799: thank you for the feedback, @Vukkyy. I've gone ahead and updated the error message.

seniorquico (approved):

PR #789, opened 2021-09-09
Show Autostart Page if Autostart Not Enabled

jpark799:
If users have not enabled auto-start during their onboarding process, we will show the auto-start page again on startup but with a 'Do Not Show Again' checkbox. 

We only want to show this page again with the do not show again checkbox if users 
- have not seen the auto-start page before
- are on the desktop app
- don't have auto-start enabled
- haven't checked 'do not show me again'

### DIFF
--- a/packages/web-app/src/Store.ts
+++ b/packages/web-app/src/Store.ts
@@ -175,6 +175,9 @@ export class RootStore {
       ])
 
       this.finishInitialLoading()
+      if (this.onboarding.shouldShowAutoStartPageAgain) {
+        this.onboarding.showAutoStartPageAgain()
+      }
       this.onboarding.showOnboardingIfNeeded()
     }.bind(this),
   )
--- a/packages/web-app/src/Store.ts
+++ b/packages/web-app/src/Store.ts
@@ -177,10 +177,8 @@ export class RootStore {
       ])
 
       this.finishInitialLoading()
-      if (this.onboardingAutoStart.shouldShowAutoStartPageAgain) {
-        this.onboardingAutoStart.showAutoStartPageAgain()
-      }
       this.onboarding.showOnboardingIfNeeded()
+      this.onboarding.reshowOnboardingPagesIfNeeded()
     }.bind(this),
   )
 
--- a/packages/web-app/src/Store.ts
+++ b/packages/web-app/src/Store.ts
@@ -11,7 +11,7 @@ import { EngagementStore } from './modules/engagement'
 import { HomeStore } from './modules/home/HomeStore'
 import { AutoStartStore, MachineSettingsUIStore, MachineStore, NativeStore } from './modules/machine'
 import { NotificationStore } from './modules/notifications'
-import { OnboardingAntivirusStore, OnboardingStore } from './modules/onboarding'
+import { OnboardingAntivirusStore, OnboardingAutoStartStore, OnboardingStore } from './modules/onboarding'
 import { ProfileStore } from './modules/profile'
 import { Profile } from './modules/profile/models'
 import { ReferralStore } from './modules/referral'
@@ -76,6 +76,7 @@ export class RootStore {
   public readonly bonuses: BonusStore
   public readonly seasons: SeasonsStore
   public readonly onboarding: OnboardingStore
+  public readonly onboardingAutoStart: OnboardingAutoStartStore
   public readonly onboardingAntivirus: OnboardingAntivirusStore
   public readonly saladFork: SaladFork
   public readonly startButtonUI: StartButtonUIStore
@@ -113,6 +114,7 @@ export class RootStore {
     this.bonuses = new BonusStore(this, axios)
     this.seasons = new SeasonsStore(axios)
     this.onboarding = new OnboardingStore(this)
+    this.onboardingAutoStart = new OnboardingAutoStartStore(this)
     this.onboardingAntivirus = new OnboardingAntivirusStore(this)
     this.startButtonUI = new StartButtonUIStore(this)
     this.machineSettingsUI = new MachineSettingsUIStore(this)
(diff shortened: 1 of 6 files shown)


packages/web-app/src/modules/onboarding/OnboardingStore.tsx
```diff
@@ -21,75 +21,75 @@ export class OnboardingStore {
...
 
   public disableSleepModePending: boolean = false
 
   @observable
   public disableSleepModeErrorMessage?: string = undefined
 
-  @observable
-  public enableAutoStartPending: boolean = false
+  @computed
+  public get hasCompletedOnboarding(): boolean {
```
jpark799, line 63: rename to hasCompletedAvailableOnboardingPages

packages/web-app/src/modules/onboarding/OnboardingStore.tsx
```diff
@@ -21,75 +21,75 @@ export class OnboardingStore {
...
+    // if this is true, then we can push users to show autostartpageagain. but will maybe(?) have to have this above the showonboarding pages.
+    const completedOnboarding =
+      this.onboardingPagesCompleted !== null &&
+      this.availableOnboardingPages.length <= JSON.parse(this.onboardingPagesCompleted).length
```
jpark799, line 67: do an exact check here instead of just checking the length

packages/web-app/src/modules/onboarding/OnboardingAutoStartStore.tsx
```diff
@@ -0,0 +1,78 @@
...
+  public enableAutoStartErrorMessage?: string = undefined
+
+  @observable
+  public enableAutoStartPending: boolean = false
+
+  @observable
+  public isDoNotShowAutoStartAgainChecked: boolean = false
+
+  @computed
+  public get haveSeenAutoStartPage(): boolean {
```
jpark799, line 19: renamed to hasSeenAutoStartPage

packages/web-app/src/Store.ts
```diff
@@ -175,6 +177,9 @@ export class RootStore {
       ])
 
       this.finishInitialLoading()
+      if (this.onboardingAutoStart.shouldShowAutoStartPageAgain) {
```
scottcinatl, line 180: We will most likely be adding more of these little nudges to change a setting that was previously set in onboarding. So it might be helpful to provide a more generic pattern for showing these pages to the user.

packages/web-app/src/modules/onboarding-views/pages/AutoStartConfigurationPage.tsx
```diff
@@ -80,13 +80,19 @@ export interface AutoStartConfigurationPageProps extends WithStyles<typeof style
   enableAutoStartErrorMessage?: string
   onEnableAutoStart: () => void
   onSkipAutoStart: () => void
+  haveSeenAutoStartPage: boolean
+  isDoNotShowAutoStartAgainChecked: boolean
```
scottcinatl, line 84: nit: change `isDoNotShowAutoStartAgainChecked` to `doNotShowAutoStartAgain`

packages/web-app/src/modules/onboarding-views/pages/AutoStartConfigurationPage.tsx
```diff
@@ -127,6 +133,13 @@ const _AutoStartConfigurationPage = ({
                 />
               </span>
             </div>
+            {haveSeenAutoStartPage && (
```
scottcinatl, line 136: nit: change `haveSeenAutoStartPage` to `hasSeenAutoStartPage`

packages/web-app/src/modules/onboarding/OnboardingAutoStartStore.tsx
```diff
@@ -0,0 +1,78 @@
...
+      this.store.onboarding.onboardingPagesCompleted !== null &&
+      this.store.onboarding.hasCompletedOnboarding &&
+      !this.userHasSelectedDoNotShowAutoStartAgain &&
+      this.store.onboarding.onboardingPagesCompleted.includes(ONBOARDING_PAGE_NAMES.AUTO_START_CONFIGURATION)
+    return shouldShowAutoStartPageAgain
+  }
+
+  @computed
+  public get userHasSelectedDoNotShowAutoStartAgain(): boolean {
+    if (Storage.getItem(DO_NOT_SHOW_AUTO_START_AGAIN)) {
```
scottcinatl, line 40: I believe this storage method returns a string. If we are wanting to return a boolean based on the storage item value, we should probably be more explicit about what we determing as the right condition for `true`.

packages/web-app/src/modules/onboarding/OnboardingAutoStartStore.tsx
```diff
@@ -0,0 +1,78 @@
...
+    this.enableAutoStartErrorMessage = undefined
+    this.enableAutoStartPending = true
+    try {
+      yield this.store.autoStart.setAutoStart(true)
+    } catch {
+      this.enableAutoStartErrorMessage =
+        'Something went wrong and we were unable to adjust your auto start settings. You can adjust these settings yourself in the Desktop App Settings, or contact support for assistance.'
+    } finally {
+      this.enableAutoStartPending = false
+      this.store.onboarding.viewNextPage(ONBOARDING_PAGE_NAMES.AUTO_START_CONFIGURATION)
```
scottcinatl, line 60: We shouldn't be mixing the action of changing a user's settings with a routing decision. There should be a separate action that calls `enableAutoStart` and then provide whatever other functionality desired after, in this case, a routing action.

scottcinatl (changes requested):

scottcinatl (approved):

PR #803, opened 2021-10-01
Remove logic that defaults subheading to "free"

scottcinatl:


seniorquico (approved):

seniorquico:
Does this change mean swag will no longer be free?

scottcinatl:
> 
> 
> Does this change mean swag will no longer be free?

Yes, it was a tough decision. But it had to be done ;)

PR #832, opened 2021-11-06
If user has reached last xp level default to 100% complete

vtheskeleton:
When the user reaches the last Xp milestone, the display on the right side of the screen should remain at the last milestone instead of getting more and more scrambled again.
This should default the completion to 100% if the user has more Xp than the milestone, but honestly i have no clue if it actually works

### DIFF
--- a/packages/web-app/src/modules/xp/ExperienceStore.ts
+++ b/packages/web-app/src/modules/xp/ExperienceStore.ts
@@ -21,6 +21,7 @@ export class ExperienceStore {
   }
 
   @computed get currentPercentComplete(): number {
+    if (this.currentXp > 131400) return 1;
     const level = this.currentLevel
     let totalRange = level.maxXp - level.minXp
     let delta = this.currentXp - level.minXp


seniorquico:
The suggested solution won't be correct. From season to season, there will be (minor) differences in the total XP required to reach the last level.

I'm not clear on what problem this is trying to solve. Could you elaborate on the issue?

vtheskeleton:
> The suggested solution won't be correct. From season to season, there will be (minor) differences in the total XP required to reach the last level.
> 
> I'm not clear on what problem this is trying to solve. Could you elaborate on the issue?

Ah crap, thought that was for the pantry but now that you mention that I am realizing that probably isnt the case

vtheskeleton:
image
The actual issue im trying to solve is this, after reaching the last item in the pantry the thing on the right starts to become a mess

jpark799:
Thank you for bringing this to our attention! We've gone ahead and implemented the change with [this PR](https://github.com/SaladTechnologies/salad-applications/pull/834)

PR #834, opened 2021-11-12
Fix Pantry Animation Bug

jpark799:
If a user has reached the XP level required to get all the vegies(131400) in the pantry, the animation will stop moving.


scottcinatl (approved):

Vukkyy:
nit: the variable says "vegies"?

PR #843, opened 2021-11-30
Reset account onboarding pages completed on logout

scottcinatl:
Since all onboarding pages are tracked via local storage, if a different user logs into the Salad App on a friends computer for the first time, they will never see a welcome or referral code onboarding page. Resetting these account onboarding pages on logout will allow for a different user to setup their account in this scenario.

### DIFF
--- a/packages/web-app/src/modules/onboarding/OnboardingStore.tsx
+++ b/packages/web-app/src/modules/onboarding/OnboardingStore.tsx
@@ -168,6 +168,28 @@ export class OnboardingStore {
     }
   }
 
+  /**
+   * Removes all account related onboarding pages from local storage.
+   */
+  @action
+  public resetAccountOnboardingPagesCompleted = () => {
+    if (this.onboardingPagesCompleted) {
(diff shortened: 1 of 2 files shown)


packages/web-app/src/modules/onboarding/OnboardingStore.tsx
```diff
@@ -168,6 +168,28 @@ export class OnboardingStore {
     }
   }
 
+  /**
+   * Removes all account related onboarding pages from local storage.
+   */
+  @action
+  public resetAccountOnboardingPagesCompleted = () => {
```
jpark799, line 175: can we rename this action and the comment above to indicate that only the welcome and referral pages are being reset? with the current name it implied the entire onboarding process is being reset

jpark799: okay, that sounds fair. i was thinking of the entire process as just one onboarding process but you're definitely right that we can make a distinction between account onboarding and machine onboarding.

jpark799 (approved):

PR #853, opened 2022-01-10
install saladbowl and saladbowl runner

thesaladmander:
Updates the installer to install the Salad.Bootstrapper into Salad/SaladBowl and the previous SaladBowl into Salad/SaladBowlRunner.  I also update the appsettings file for the bootstrapper to point to the installed SaladBowlRunner/Salad.Bowl.Service.exe.  So the appsettings file is assumed to have an SBExe and SBDir in the configuration to be replaced by the appropriate location for the runner.

The result is that the installer should run 'SaladBowl' (now referencing Salad.Boostrapper) on Session 0 which will run SaladBowlRunner on Session 1 (or something non-0).

Kyle: I removed the underscores on the install files.  Did you want me to check those in that way or not?


packages/desktop-app/build/salad-bowl/_install.nsh
```diff
@@ -18,16 +18,26 @@ ${EndIf}
 
 RMDir /r "$LocalAppData\Salad\workload-definitions\gpuz"
 RMDir /r "$LocalAppData\Salad\workload-definitions\systeminformation"
+
+
 SetOutPath "$INSTDIR\SaladBowl"
-File /r "${__FILEDIR__}\dist\service\*.*"
+File /r "${__FILEDIR__}\dist\service\SaladBowl\*.*"
+!insertmacro StrRep $7 $INSTDIR "\" "\\"  ; need to escape the slashes or dotnet will be unhappy
+!insertmacro _ReplaceInFile "$INSTDIR\SaladBowl\appsettings.json" "SBEXE" "$7\\SaladBowlRunner\\Salad.Bowl.Service.exe"
```
seniorquico, line 26: Let's move the path to `Salad.Bowl.Service.exe` to the service's command line arguments.

seniorquico (changes requested):

packages/desktop-app/build/salad-bowl/_install.nsh
```diff
@@ -18,16 +18,22 @@ ${EndIf}
...
+
 SetOutPath "$LocalAppData\Salad\workload-definitions"
 File /r "${__FILEDIR__}\dist\workload-definitions\*.*"
 SetOutPath "$LocalAppData\Salad\workloads"
 File /r "${__FILEDIR__}\dist\workloads\*.*"
 SetOutPath $INSTDIR
 
 install_service_retry:
-  SimpleSC::InstallService "SaladBowl" "Salad Bowl" "16" "2" "$INSTDIR\SaladBowl\Salad.Bowl.Service.exe" "" "" ""
+  SimpleSC::InstallService "SaladBowl" "Salad Bowl" "16" "2" "$INSTDIR\SaladBowl\Salad.Bootstrapper.exe" "" "" ""
```
seniorquico, line 36: According to the [SimpleSC docs](https://nsis.sourceforge.io/NSIS_Simple_Service_Plugin), the binary path should include the arguments. I also think we need to explicitly quote them to cover potential spaces in the `$INSTDIR` path.

```suggestion
  SimpleSC::InstallService "SaladBowl" "Salad Bowl" "16" "2" "$INSTDIR\SaladBowl\Salad.Bootstrapper.exe --sb $\"$INSTDIR\SaladBowlRunner\Salad.Bowl.Service.exe$\"" "" "" ""
```

packages/desktop-app/build/salad-bowl/_install.nsh
```diff
@@ -63,8 +69,10 @@ install_cancel:
   Abort
 
 install_continue:
-  SimpleSC::StartService "SaladBowl" "" 30
+  SimpleSC::StartService "SaladBowl" "--sb $INSTDIR\SaladBowlRunner\Salad.Bowl.Service.exe" 30
```
seniorquico, line 72: If the argument is baked into the service installation, then this may be unnecessary?

seniorquico (changes requested):

packages/desktop-app/build/salad-bowl/_install.nsh
thesaladmander: yeah, this wasn't working.  I did test it, but I screwed up the test.  So the trick as you'll see is the whole string needs to be in single quotes and inside you can use double quotes.  Tried the reverse initially and it didn't work.  I tried your escapes and it didn't compile.

seniorquico (approved):

PR #854, opened 2022-01-17
salad.com/___ web app

dsarfati:


packages/web-app/src/index.tsx
```diff
@@ -1,35 +1,33 @@
 // Import CSS. Order is important!
-import './index.css'
```
scottcinatl, line 2: Anytime we add anything to this file, we have to `save without formatting` in VS Code in order to keep the order of imports the same. So we should only be seeing the added imports to the file, not the re-organizing of the import order.

jpark799 (changes requested):
image
do we want Logout to be in its own column like this?

packages/web-app/src/Routes.tsx
```diff
@@ -52,53 +43,67 @@ const _Routes = ({ location }: RouteComponentProps) => {
...
+        exact
+        path="/warnings/override-compatibility-detection"
+        component={OverrideCompatibilityDetectionContainer}
+      />
+
+      {/* Account */}
+      <Redirect exact from="/account" to="/account/summary" />
+      <Route path="/account" component={SettingsContainer} />
+
+      {/* <Route exact path="/errors/cuda" component={CudaErrorContainer} />
```
scottcinatl, line 79: These error pages should not be commented out. These are the error modals that will occur in the app when a user encounters issues running Salad. If you want these to only be in the app and not `salad.com` then there will need to be some logic here to exclude these routes based on the domain.

packages/web-app/src/modules/settings-views/SettingsContainer.ts
```diff
@@ -5,45 +5,29 @@ import { RootStore } from '../../Store'
...
+    { url: '/account/bonuses', text: 'Bonuses', component: BonusPageContainer },
     {
-      url: '/settings/desktop-settings',
+      url: '/account/desktop-settings',
       text: 'Desktop App Settings',
       component: DesktopSettingsContainer,
       divider: true,
       desktopOnly: true,
     },
-    {
```
scottcinatl, line 41: Will the user not be able to access the link to `What's New` while on the app or salad.com? This is was the only link we have in the app that leads the user there.

packages/web-app/src/modules/settings-views/SettingsContainer.ts
```diff
@@ -52,7 +36,7 @@ const mapStoreToProps = (store: RootStore): any => {
   }
 
   const handleClose = () => {
-    store.analytics.trackSmartLink('/', 'Back')
+    store.analytics.trackSmartLink('/store', 'Back')
```
scottcinatl, line 39: This isn't a bug introduced by you, but this track smart link is not accurately portraying what is happening. The user is clicking back, but they are not being taken back to the '/` or `/store`. The `hideModal` method actually just returns the user back to their previous location. So data for this event are wrong.

scottcinatl (changes requested):
There is also an issue when clicking the profile name in the navigation bar. this leads the user to a route that no longer exists. 

account

scottcinatl:
@dsarfati If I change the basename to `/app` and the port to `3001` for my environment variables, I get directed to `http://localhost:3001/app/profile-error` instead of just seeing the working app but based from `/app`

I am not sure if any of your routing changes have affected this flow, for this was to test how the app would working on the `salad.com` domain.

packages/web-app/src/Routes.tsx
dsarfati: We are going to revisit these once we figure out what we can/can't do from the app connected to SB. With the newly designed errors project for the widget, Im not sure we will support starting from salad.com

seniorquico:
> @dsarfati If I change the basename to `/app` and the port to `3001` for my environment variables, I get directed to `http://localhost:3001/app/profile-error` instead of just seeing the working app but based from `/app`
> 
> I am not sure if any of your routing changes have affected this flow, for this was to test how the app would working on the `salad.com` domain.

We won't set the basename to `/app` anymore. It will remain its default (i.e. `/`).

dsarfati:
>

packages/web-app/src/Routes.tsx
scottcinatl: But these routes still need to be available for the `app.salad.io` version, right?

packages/web-app/src/index.tsx
dsarfati: Fixed by Kyle

packages/web-app/src/modules/settings-views/SettingsContainer.ts
dsarfati: We will be able to get to the what's new directly on the website so we don't need the link in here anymore. For example, the widget has a direct link off to the whats new page

dsarfati:
> 



> **A reviewer left a comment:**
> 
> recording
> 
> it looks like we're going to have to update every `/settings/summary` with `/account/summary`
> Browser metadata
> 
> [Open Deploy Preview](https://deploy-preview-854--salad-com-app.netlify.app/settings/summary) · [Mark as Resolved](https://app.netlify.com/cdp/resolve?deployID=61e5b53c9d39fc00071ddb7d&commentID=61e5be48a71e1ecfb173aeba&resolution=resolved)

I don't see this path in the garden component. I don't know why that is happening.

> image do we want Logout to be in its own column like this?

I really dont like it like this, but I liked it even less at the bottom of the other row. I think we will take this as an item to update in the future with a custom UI component.

packages/web-app/src/Routes.tsx
dsarfati: Yes, we are not pushing these new changes to app.salad.io. They will only go to salad.com. App.salad.io will now live on the `alpha` branch

scottcinatl:
@dsarfati Just ran your updated branch, and was directed to `/profile-error` on starting the app. I am not in an authenticated state, do you see this happening on your end?

profile error page

jpark799:
> > 
> 
> > **A reviewer left a comment:**
> > recording
> > it looks like we're going to have to update every `/settings/summary` with `/account/summary`
> > Browser metadata
> > [Open Deploy Preview](https://deploy-preview-854--salad-com-app.netlify.app/settings/summary) · [Mark as Resolved](https://app.netlify.com/cdp/resolve?deployID=61e5b53c9d39fc00071ddb7d&commentID=61e5be48a71e1ecfb173aeba&resolution=resolved)
> 
> I don't see this path in the garden component. I don't know why that is happening.
> 
> > image do we want Logout to be in its own column like this?
> 
> I really dont like it like this, but I liked it even less at the bottom of the other row. I think we will take this as an item to update in the future with a custom UI component.

looks like the 404 is resolved now. thanks for taking a look at that.

addressing the logout stuff in a later update sounds good. i just wanted to make sure it was something we were thinking about.

jpark799 (approved):

scottcinatl (approved):

PR #860, opened 2022-01-24
[Bug] broken Auto Start link

jpark799:
we had a link to the desktop settings page that would always show even in the webapp. this page wasn't accessible though unless you were in the desktop app so i added a check to see if we're in the native app. the path has also changed so i updated it. 
image

### DIFF
--- a/packages/web-app/src/modules/earn-views/components/MiningInformation.tsx
+++ b/packages/web-app/src/modules/earn-views/components/MiningInformation.tsx
@@ -79,13 +79,9 @@ class _MiningInformation extends Component<Props> {
 
             <P>
               Don't forget to enable{' '}
-              {isNative ? (
-                <b>
-                  <SmartLink to="/account/desktop-settings">Auto Start</SmartLink>
-                </b>
-              ) : (
-                <b> Auto Start </b>
-              )}
+              <b>
+                {isNative ? <SmartLink to="/account/desktop-settings">Auto Start</SmartLink> : <span>Auto Start</span>}
+              </b>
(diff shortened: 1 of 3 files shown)


packages/web-app/src/modules/earn-views/components/MiningInformation.tsx
```diff
@@ -77,9 +79,13 @@ class _MiningInformation extends Component<Props> {
 
             <P>
               Don't forget to enable{' '}
-              <b>
-                <SmartLink to="/settings/desktop-settings">Auto Start</SmartLink>
-              </b>
+              {isNative ? (
+                <b>
```
scottcinatl, line 83: I would remove the blank spaces from your `Auto Start` text block that doesn't include the link. 

nit: you could just use the ternary inside the the html tag. 

```
  <b>
       {isNative ? <SmartLink to="/account/desktop-settings">Auto Start</SmartLink> : "Auto Start"}
  </b>
```

scottcinatl (changes requested):
Does a similar change need to be made in the alpha branch as well? To at least remove the link when the user is on the browser instead of the desktop app?

packages/web-app/src/modules/earn-views/components/MiningInformation.tsx
jpark799: yeah, we're going to need to make the change to the alpha project as well. once this PR is approved, i can make the change in that branch.

great call on making the ternary less messy. i like that way more and i'll get that implemented.

seniorquico (approved):

PR #861, opened 2022-01-24
[Bug] Open Earn FAQ links in same window

jpark799:
We had a couple of links that were opening in new windows but we wanted them to open in the same window. 

I'm not super in love with the way I did this but because the base url is different, I couldn't just do a `to="/article"` which would open in the same window. I changed our `SmartLink` component to have an optional prop called `openInSameWindow` which removes the `target='_blank'`. This doesn't change the behavior of any of our existing `SmartLink` component uses.

Please let me know if there's another way you would prefer for this to work.

### DIFF
--- a/packages/web-app/src/components/SmartLink.tsx
+++ b/packages/web-app/src/components/SmartLink.tsx
@@ -23,7 +23,6 @@ interface Props extends WithStyles<typeof styles> {
   children?: ReactNode
   className?: string
   trackingInfo?: LinkTrackingInfo
-  openInSameWindow?: boolean
 }
 
 const handleClickTracking = (to?: string, trackingInfo?: any) => {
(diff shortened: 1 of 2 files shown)


packages/web-app/src/components/SmartLink.tsx
```diff
@@ -41,7 +42,7 @@ const _SmartLink = ({ to, children, classes, className, trackingInfo }: Props) =
       <a
         className={finalClassName}
         href={to}
-        target={'_blank'}
+        target={openInSameWindow ? undefined : '_blank'}
```
seniorquico, line 45: Is there ever a valid use case to open in `target="_blank"`?

jpark799: yeah, there are some cases where we're linking to our subreddit, discord, other antivirus websites, trustpilot, and things like that.

jpark799:
we're only opening SmartLinks in a new window if we're in the native app. If not, we will always open in the link in the same window due to accessibility.

seniorquico (approved):

PR #880, opened 2022-02-03
Updates Season Page

jpark799:
Changed the title of the Seasons Page from Current Season to Last Season Snapshot and hid the time estimates on level cards

image


seniorquico (changes requested):
This breaks the nav bar in the alpha Electron app.

seniorquico (approved):

PR #881, opened 2022-02-03
updated seasons page and pulled in alpha garden

jpark799:
pulling in the changes to the seasons page while not breaking the navbar

image

navbar 
gif


seniorquico (changes requested):
This should also remove the `lib/...-1.0.66.tgz` file.

seniorquico (approved):

PR #882, opened 2022-02-05
Updated Trex & Xmrig

mezotv:
Updates Trex and Xmrig to the newest version avalible.

### DIFF
--- a/packages/web-app/src/modules/salad-bowl/definitions/trex.tsx
+++ b/packages/web-app/src/modules/salad-bowl/definitions/trex.tsx
@@ -2,6 +2,11 @@ const baseUrl = 'https://github.com/SaladTechnologies/plugin-downloads/releases/
 
 export const downloads = [
   {
+    version: '0.25.2',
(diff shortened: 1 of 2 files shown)


seniorquico:
Thanks, @mezotv. Unfortunately, there are a couple issues.

First, T-Rex miner isn't currently used from the Electron app due to virtual terminal incompatibilities. We're working on a solution, but I don't have a definite timeline. I'd guess it might be ready in a few more weeks.

Second, we can't update these plugin definitions without also publishing the needed binaries. The versions/URLs you're referencing aren't yet available.

@MerlynTheWizard is working on publishing new packages and updating the plugin definitions. I'll get this over to him, and he can at least work on getting the latest XMrig released.

Right now, it's tough to accept contributions to the plugin definitions. However, I'd like to encourage you to stay tuned! We're going to be creating a new, open source GitHub repo in the coming weeks where we'll fix these workflow issues and give Chefs direct access to our automated publishing pipelines!

PR #896, opened 2022-03-05
Changed page to recipe on error page

mezotv:
Some user in the salad discord submitted this as a feedback i found it cool so here it is. I belive this adds some extra "salad" flair to the site!


FemboyNova:
This thing is good

seniorquico:
Thanks for the fun recommendation. We've decided to keep it as-is to minimize confusion for newcomers not accustomed to our puns.

PR #901, opened 2022-03-15
AV helpscout articles

jpark799:
Just a quick fix for making sure our AV guides are being displayed for now. I only touched what I needed to get this fix out the door. We're going to need to eventually come back and rip out all this zendesk stuff and come up with a more permanent solution.  

From what I could tell after checking the routes, the only articles we were displaying from zendesk were the AV ones.

image
http://localhost:3000/onboarding/antivirus-guide/360033124692
image
http://localhost:3000/onboarding/antivirus-guide/360042153731

### DIFF
--- a/packages/web-app/src/modules/onboarding-views/pages/AntivirusGuide.tsx
+++ b/packages/web-app/src/modules/onboarding-views/pages/AntivirusGuide.tsx
@@ -213,7 +213,6 @@ class _AntivirusGuide extends Component<AntivirusGuideProps, State> {
         <OnboardingAntiVirusScrollbar>
           <div className={classes.page}>
             <div className={classes.container}>
-              {console.log(helpScoutUrl)}
               <div className={classes.heading}>
                 <div className={classes.title}>
                   <Text variant="headline">{antivirusName}</Text>
--- a/packages/web-app/src/modules/onboarding-views/pages/AntivirusGuide.tsx
+++ b/packages/web-app/src/modules/onboarding-views/pages/AntivirusGuide.tsx
@@ -3,7 +3,6 @@ import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
 import { Button, Modal, Text } from '@saladtechnologies/garden-components'
 import { Component } from 'react'
 import withStyles, { WithStyles } from 'react-jss'
-import Skeleton from 'react-loading-skeleton'
 import { ModalPage } from '../../../components'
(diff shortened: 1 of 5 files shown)


packages/web-app/src/modules/onboarding-views/pages/AntivirusGuide.tsx
```diff
@@ -220,48 +213,26 @@ class _AntivirusGuide extends Component<AntivirusGuideProps, State> {
...
-                        </div>
-                      )}
-                      <Text variant="baseL">
-                        <div className={classes.content} dangerouslySetInnerHTML={{ __html: article }} />{' '}
-                      </Text>
-                    </>
-                  ) : null}
-                  <span className={classes.closeButtonContainer}>
-                    <Button size="medium" label="Close" onClick={onCloseClicked} />
+              {console.log(helpScoutUrl)}
```
scottcinatl: remove console.log

packages/web-app/src/modules/onboarding-views/pages/AntivirusGuide.tsx
```diff
@@ -220,48 +213,26 @@ class _AntivirusGuide extends Component<AntivirusGuideProps, State> {
...
+                <div className={classes.title}>
+                  <Text variant="headline">{antivirusName}</Text>
+                </div>
+              </div>
+
+              <iframe width={700} height={700} title={antivirusName} src={helpScoutUrl} />
+              <div className={classes.subtitle}>
+                <Text variant="baseS">
+                  Use a different anti-virus provider?{' '}
+                  <span className={classes.link} onClick={this.handleOpenAVSelectionModal}>
```
scottcinatl, line 226: What happens they click this button to show the list of Articles? I assume it still open to a page of links that don't exist anymore. We are going to have to change that as well.

scottcinatl (changes requested):
It seems that you have the individual pages covered on the onboarding flow. What still is needed is to correct (somehow) the list of available AV guide links page. 

Didn't we also have another way to get to these AV pages, through an error route? If so, I think those pages still need to be addressed.

packages/web-app/src/modules/onboarding-views/pages/AntivirusGuide.tsx
jpark799: we're building the list of articles in the react app and not zendesk so it still displays as usual. 


av gif

scottcinatl (approved):

PR #902, opened 2022-03-15
Loading Helpscout AV articles

jpark799:
the onboarding pages will be the same but in our alpha branch, we are using the `error/` routes. of the error routes, only the three below were pulling from zendesk

**- `/errors/anti-virus/:id`**
image
image
clicking the link will bring us to the generic antivirus route.
av gif2

**- `/errors/anti-virus/`** 
image

**- `/errors/firewall`**
image

### DIFF
--- a/packages/web-app/src/modules/error-views/components/AntiVirusFirewallErrorPage.tsx
+++ b/packages/web-app/src/modules/error-views/components/AntiVirusFirewallErrorPage.tsx
@@ -1,12 +1,12 @@
 import { faTimes } from '@fortawesome/free-solid-svg-icons'
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
-import type { ReactChild } from 'react'
 import { Component } from 'react'
 import withStyles, { WithStyles } from 'react-jss'
 import Skeleton from 'react-loading-skeleton'
 import { Button, InfoButton, ModalPage, SmartLink } from '../../../components'
 import { SaladTheme } from '../../../SaladTheme'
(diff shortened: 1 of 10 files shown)


packages/web-app/src/modules/error-views/components/AntiVirusFirewallErrorPage.tsx
```diff
@@ -189,14 +192,13 @@ class _AntiVirusFirewallErrorPage extends Component<Props, State> {
     const {
       antivirusName,
       errorType,
-      article,
-      articleList,
       fallthrough,
       loading,
       onCloseClicked,
-      onViewArticle,
```
scottcinatl, line 197: This action to view article didn't just navigate a user to the correct page, it also had business logic that tracked which article a user clicked on for mixpanel. So these clicks should be captured just as they were before.

packages/web-app/src/modules/error-views/components/AntiVirusFirewallErrorPage.tsx
```diff
@@ -189,14 +192,13 @@ class _AntiVirusFirewallErrorPage extends Component<Props, State> {
     const {
       antivirusName,
       errorType,
-      article,
```
scottcinatl, line 192: If we are removing props from being needed, I would expect that their would be changes made in the container components as well that passes down those props.

scottcinatl (changes requested):

jpark799:
new general AV page. 
image

scottcinatl (approved):

PR #912, opened 2022-04-28
[App] Adding redirect_uri on login page

jpark799:
if you click on any of the login buttons in the app, we will pass the path to the `redirect_uri` which will redirect you back to the page after authenticating.

### DIFF
--- a/packages/web-app/src/modules/auth/AuthStore.ts
+++ b/packages/web-app/src/modules/auth/AuthStore.ts
@@ -101,8 +101,13 @@ export class AuthStore {
 
       // Save current route so we can go back to it later once it is resolved
       this.startingRoute = this.router.location.pathname === '/login' ? '/store' : this.router.location.pathname
+      const path = window.location.pathname
 
-      this.router.replace('/login')
+      if (path === '/') {
+        this.router.replace(`/login`)
+      } else {
+        this.router.replace(`/login/?redirect_uri=${path}`)
+      }
     })
 
     await this.loginPromise
--- a/packages/web-app/src/modules/auth/AuthStore.ts
+++ b/packages/web-app/src/modules/auth/AuthStore.ts
@@ -41,11 +41,6 @@ export class AuthStore {
 
   private loginPromise?: Promise<void>
 
-  /**
-   * Path to where we started the login process.
-   */
-  private startingRoute?: string
-
   private loginResolve?: () => void
   private loginReject?: () => void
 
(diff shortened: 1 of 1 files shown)


packages/web-app/src/modules/auth/AuthStore.ts
```diff
@@ -101,8 +101,13 @@ export class AuthStore {
 
       // Save current route so we can go back to it later once it is resolved
       this.startingRoute = this.router.location.pathname === '/login' ? '/store' : this.router.location.pathname
+      const path = window.location.pathname
 
-      this.router.replace('/login')
+      if (path === '/') {
+        this.router.replace(`/login`)
+      } else {
+        this.router.replace(`/login/?redirect_uri=${path}`)
```
seniorquico, line 109: But this will navigate to the app's _internal_ login page. That page doesn't support `redirect_uri`. We'd need to intercept `/login` navigation events and explicitly push them to the _external_ login app.

seniorquico (changes requested):

packages/web-app/src/modules/auth/AuthStore.ts
```diff
@@ -101,8 +101,13 @@ export class AuthStore {
 
       // Save current route so we can go back to it later once it is resolved
       this.startingRoute = this.router.location.pathname === '/login' ? '/store' : this.router.location.pathname
```
scottcinatl, line 103: Is `startingRoute` relevant anymore? If Login is now handling the redirect?

scottcinatl (changes requested):

packages/web-app/src/modules/auth/AuthStore.ts
jpark799: oh great catch. i guess it isn't

packages/web-app/src/modules/auth/AuthStore.ts
jpark799: aw actually `window.location.replace` should work. lmk if there's another method that you would prefer though

seniorquico: Hard-coding it to `salad.com` will break non-production environments, though. It's best if it's kept as an absolute _path_ (i.e. `/login?redirect_uri=...`). You just need to explicitly break out of React Router navigation and let the browser leave/go to the new login app.

seniorquico (approved):

PR #921, opened 2022-05-09
Need help with running a local build of the web-app

UltimatePrograms:
How do I run web-app?
I can't get past Module not found: Can't resolve '@saladtechnologies/garden-components'
And I think it must be https://pkgs.dev.azure.com/SaladTechnologies/_packaging/SaladTechnologies/npm/registry/


FemboyNova:
From looking into it some of the modules are private so we cant run it

PR #925, opened 2022-05-16
[SaladCard] make sure user is enrolled in SaladCard before redeeming SaladCard rewards

jpark799:
Like with Minecraft and Paypal, we're now making sure users have a SaladCard before allowing them to redeem SaladCard rewards. 

If a user doesn't have a SaladCard, they'll see this red link on SaladCard rewards. 
image

If users somehow do get around this check and redeem the reward without having a SaladCard, they will see this error notification. the copy will likely change later but we wanted to get a working mvp in for now.
image

clicking the notification or link will take users to the enroll saladcard page

### DIFF
--- a/packages/web-app/src/modules/reward-views/pages/RewardDetailsPage.tsx
+++ b/packages/web-app/src/modules/reward-views/pages/RewardDetailsPage.tsx
@@ -42,6 +42,7 @@ interface Props extends WithStyles<typeof styles> {
   requiresMinecraftUsername: boolean
   trackDisabledBuyNowClick: () => void
   requiresPayPalAccount: boolean
+  requiresSaladCard: boolean
 }
 
 class _RewardDetailsPage extends Component<Props> {
@@ -57,6 +58,7 @@ class _RewardDetailsPage extends Component<Props> {
       onBack,
       requiresMinecraftUsername,
       requiresPayPalAccount,
+      requiresSaladCard,
       trackDisabledBuyNowClick,
       classes,
       ...rest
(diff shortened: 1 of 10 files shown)


packages/web-app/src/modules/reward-views/pages/RewardDetailsPage.tsx
```diff
@@ -70,6 +72,7 @@ class _RewardDetailsPage extends Component<Props> {
           onRedeem={onRedeem}
           requiresMinecraftUsername={requiresMinecraftUsername}
           requiresPayPalAccount={requiresPayPalAccount}
+          requiresSaladCard={true}
```
scottcinatl, line 75: I assume this should not be hardcoded?

packages/web-app/src/modules/reward/RewardStore.ts
```diff
@@ -78,12 +81,25 @@ export class RewardStore {
         type: 'error',
       })
     }
+
+    if (requiresSaladCard) {
+      this.requiresFurtherAction = true
+      this.store.notifications.sendNotification({
+        category: NotificationMessageCategory.FurtherActionRequired,
+        title: 'You must activate a SaladCard for this reward.',
+        message: 'Click to go to the enroll a SaladCard page.',
```
scottcinatl, line 90: I know this copy is not finalized, but maybe `Click to enroll for a SaladCard` might be better.

scottcinatl (changes requested):

scottcinatl (approved):

PR #927, opened 2022-05-24
[SaladCard] Completed first iteration of SaladCard details page

jpark799:
image

the saladcard image is just a placeholder for now. 
have added the ability to view SaladCard balance, lock card, and replace.


packages/web-app/src/Routes.tsx
```diff
@@ -94,7 +95,8 @@ const _Routes = ({ location }: RouteComponentProps) => {
       {saladBowlEnabled && <Redirect exact from="/settings/desktop-settings" to="/earn/machine-settings" />}
 
       <Redirect exact from="/earn" to="/earn/summary" />
-      {saladBowlEnabled && <Route path="/earn/saladcard" component={SaladCardEnrollmentPageContainer} />}
+      {saladBowlEnabled && <Route path="/earn/saladcard-enroll" component={SaladCardEnrollmentPageContainer} />}
+      {saladBowlEnabled && <Route path="/earn/saladcard-details" component={SaladCardDetailsPageContainer} />}
```
scottcinatl, line 99: If I remember correctly, `saladBowlEnabled` is a feature flag check, should we have a different feature flag for this Salad Card work?

We could have a follow up PR fix this if it takes more work.

packages/web-app/src/modules/reward-views/components/RewardHeaderBar.tsx
```diff
@@ -188,7 +188,7 @@ class _RewardHeaderBar extends Component<Props> {
               )}
               {requiresSaladCard && authenticated && (
                 <div className={classnames(classes.priceText, classes.stockLabel, classes.insufficientBalanceLabel)}>
-                  <SmartLink to="/earn/saladcard">Enroll in SaladCard</SmartLink>
+                  <SmartLink to="/earn/saladcard-enroll">Enroll in SaladCard</SmartLink>
```
scottcinatl, line 191: Not a request for this PR, just something to think about for a future refactor job for us. Maybe we should these path names into an enum so we don't have to worry about accidently setting the wrong path all over the place, and if ever need to change the path it is all in one place.

packages/web-app/src/modules/salad-card-views/SaladCardDetailsPageContainer.tsx
```diff
@@ -0,0 +1,19 @@
+import { connect } from '../../connect'
+import { RootStore } from '../../Store'
+import { SaladCardDetailsPage } from './pages/SaladCardDetailsPage'
+
+const mapStoreToProps = (store: RootStore): any => ({
```
scottcinatl, line 5: I know we don't have a real smoothed out pattern on this web app for containers, but I think we could be typing this `mapStoreToProps` function better instead of just saying it returns `any`. If you export the component's props that you are wrapping, you could get accurate types.

`(store: RootStore): Omit<SaladCardDetailsPageProps, 'classes'>`

packages/web-app/src/modules/salad-card-views/pages/SaladCardDetailsPage.tsx
```diff
@@ -0,0 +1,114 @@
...
+import saladCardPlaceHolderImage from '../assets/SaladCardPlaceHolderImage.png'
+
+const styles = (theme: SaladTheme) => ({
+  page: {
+    backgroundImage: 'linear-gradient(to right, #56A431 , #AACF40)',
+    color: theme.darkBlue,
+    flex: 1,
+  },
+  parentContainer: {
+    width: 688,
```
scottcinatl, line 20: Should this be `maxWidth: 688, width: '100%'`?

packages/web-app/src/modules/salad-card/SaladCardStore.ts
```diff
@@ -57,5 +72,64 @@ export class SaladCardStore {
     }
   })
 
+  @action.bound
+  public loadSaladCardBalance = flow(function* (this: SaladCardStore) {
+    try {
+      const response = yield this.axios.get(`/api/v2/salad-card/cards/${this.saladCard?.cardId}/balance`)
+
+      this.saladCardBalance = response.data.currentBalance
+    } catch (e) {
```
scottcinatl, line 81: Is there only one possible error response code we get here? Can we distinguish between issues of communicating with the API versus the API telling us that `cardId` doesn't exist? 

Also, will users have multiple cards? or only one?

packages/web-app/src/modules/salad-card/SaladCardStore.ts
```diff
@@ -57,5 +72,64 @@ export class SaladCardStore {
...
+      this.hasSaladCard = false
+    }
+  })
+
+  @action.bound
+  public lockSaladCard = flow(function* (this: SaladCardStore) {
+    try {
+      this.isLockSaladCardLoading = true
+      const request = true
```
scottcinatl, line 90: nit: no need to set this as a variable since it is only referenced in one place.

scottcinatl (changes requested):

packages/web-app/src/Routes.tsx
jpark799: i can double check with kyle later about getting another feature flag later as i believe he was the one that got them set up. it might be more trouble then it's worth though.

packages/web-app/src/modules/reward-views/components/RewardHeaderBar.tsx
jpark799: yeah, that's probably a good idea. only having to change it in one place would be nice.

packages/web-app/src/modules/salad-card-views/SaladCardDetailsPageContainer.tsx
jpark799: oh great catch. i'll make sure any new pages i work on use this pattern from now on.

packages/web-app/src/modules/salad-card/SaladCardStore.ts
jpark799: yeah, there's only one possible error code.

i think multiple cards may be allowed later but for now, we are working off the assumption of only having one card.

scottcinatl (approved):

PR #935, opened 2022-06-20
[SaladCard] protected view

jpark799:
I realized after last week's demo that we have access to the last 4 digits of the card so I went ahead and made the protected view of the salad card. this is what will always be viewable for users. i believe the plan is to later have a button or something that will reveal all the details of the card. 

image


packages/web-app/src/modules/salad-card/SaladCardStore.ts
```diff
@@ -115,6 +118,7 @@ export class SaladCardStore {
       this.isReplaceSaladCardLoading = true
       const response = yield this.axios.post(`/api/v2/salad-card/cards/${this.saladCard?.cardId}/replace`)
       this.saladCard = response.data
+      this.loadSaladCard()
```
scottcinatl, line 121: I am not familiar with the API, so why after we replace a salad card do we call `loadSaladCard` again? It seems that we update state of `this.saladCard` with the data we get back from the replace call. Does the `loadSaladCard` then just make an API call that returns that same data again?

packages/web-app/src/modules/salad-card-views/pages/SaladCardDetailsPage.tsx
```diff
@@ -77,7 +78,7 @@ const _SaladCardDetailsPage = ({
           <Head title="SaladCard" />
           <div className={classes.parentContainer}>
             <div>
-              <Img src={saladCardPlaceHolderImage} alt="Salad Card Image" />
+              <SaladCardProtectedView lastFourSaladCardDigits={lastFourSaladCardDigits} />
```
scottcinatl, line 81: Your types show that there may be instances where we don't have the `lastFourSaladCardDigits` what do we show the user in that case?

scottcinatl (changes requested):

packages/web-app/src/modules/salad-card/SaladCardStore.ts
jpark799: the `loadSaladCard` was to assign the `this.lastFourSaladCardDigits`. but as i read over this, i realized we didn't need to assign the response to `this.saladCard` and we can just call `this.loadSaladCard()` which will take care of both assigning `this.saladCard` and `this.lastFourSaladCardDigits`.

great catch :)

packages/web-app/src/modules/salad-card-views/pages/SaladCardDetailsPage.tsx
jpark799: just updated this to be more safe. if we don't have the `lastFourSaladCardDigits` we're going to be defaulting to 'xxxx' in the UI. if we don't have a SaladCard though, we'll be automatically redirecting users to the store and this page won't be viewable.

scottcinatl (approved):

PR #938, opened 2022-06-27
[SaladCard] Full SaladCard View

jpark799:


packages/web-app/src/modules/salad-card-views/pages/SaladCardDetailsPage.tsx
```diff
@@ -65,12 +70,25 @@ const _SaladCardDetailsPage = ({
...
+  }, [handleLoadSaladCard, handleLoadSaladBalance, handleRouteToStore, hasSaladCard, handleLoadSaladCardInfoUrl])
+
+  const [viewFullSaladCard, setViewFullSaladCard] = useState<boolean>(false)
+
+  const FiveMinutesInMs = 300000
+
+  const handleOnViewSaladCardModalClick = () => {
+    setViewFullSaladCard(true)
+    handleLoadSaladCardInfoUrl()
+    setInterval(() => setViewFullSaladCard(false), FiveMinutesInMs)
```
scottcinatl, line 89: This interval needs to be stored as a variable and cleared when this component is unmounted

packages/web-app/src/modules/salad-card/SaladCardStore.ts
```diff
@@ -82,7 +84,7 @@ export class SaladCardStore {
 
       this.saladCardBalance = response.data.currentBalance
     } catch (e) {
-      this.hasSaladCard = false
+      // this.hasSaladCard = false
```
scottcinatl, line 87: Was this meant to be commented out?

packages/web-app/src/modules/salad-card/SaladCardStore.ts
```diff
@@ -138,5 +140,15 @@ export class SaladCardStore {
     this.lastFourSaladCardDigits = this.saladCard?.panMasked.slice(-4)
   }
 
+  @action.bound
+  public loadSaladCardInfoUrl = flow(function* (this: SaladCardStore) {
+    try {
+      const response = yield this.axios.get(`/api/v2/salad-card/cards/${this.saladCard?.cardId}/embed`)
+      this.saladCardInfoUrl = response.data.url
+    } catch (e) {
+      throw e
```
scottcinatl, line 149: Is there no error UI we want to present to the user if we are unable to load this information? What does the iframe look like when this fails?

Is there any loading state that the UI needs to be aware of if this information being retrieved isn't instantly returned?

packages/web-app/src/modules/salad-card/SaladCardStore.ts
jpark799: no, my bad

packages/web-app/src/modules/salad-card/SaladCardStore.ts
jpark799: added in an error message as well as a loading spinner. good call on those

PR #940, opened 2022-08-18
Unleash / Mixpanel integration

shoneefd:
There is *probably* a race condition theoretically possible that would cause events to be lost without routing to mixpanel, but I *think* that would only happen when the AnalyticsStore crashes. Hopefully this doesn't happen often? Or if it does, hopefully it bricks the program (never thought I'd say that).

I was able to verify that this code allows the program to run as normal, but I wasn't able to verify that mixpanel will get the events. None of our current feature flags have impression data enabled, and even if they did I don't actually think I have access to the mixpanel dashboard.

### DIFF
--- a/node_modules/.yarn-integrity
+++ b/node_modules/.yarn-integrity
@@ -1,0 +1,10 @@
+{
+  "systemParams": "win32-x64-93",
+  "modulesFolders": [],
+  "flags": [],
+  "linkedModules": [],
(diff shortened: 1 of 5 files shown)


shoneefd:
The netlify checks failed because my Azure Artifacts tokens had expired. I've fixed them now in the netlify dashboard.

scottcinatl (approved):

PR #942, opened 2022-08-23
Remove zendesk references

jpark799:
We had a lot of dead code that was referencing zendesk and this PR deletes all that out.

There were some things we still needed in the zendesk store so I created a HelpScout store and moved those there.


seniorquico (changes requested):
This removes the Help Scout beacon. We need to keep this.

packages/web-app/src/modules/error-views/FallbackErrorContainer.ts
```diff
@@ -25,7 +25,6 @@ const mapStoreToProps = (store: RootStore): any => {
     alternativeMinerType,
     currentMinerType,
     onCloseClicked: () => store.ui.hideModal(true),
-    onOpenSupportTicket: () => store.zendesk.openSupportTicket(),
```
seniorquico, line 28: Shouldn't we still open the Help Scout beacon here?

packages/web-app/src/modules/error-views/components/FallbackErrorPage.tsx
```diff
@@ -31,25 +31,15 @@ interface Props extends WithStyles<typeof styles> {
   alternativeMinerType: 'GPU' | 'CPU'
   currentMinerType: 'GPU' | 'CPU'
   onCloseClicked: () => void
-  onOpenSupportTicket: () => void
```
seniorquico, line 34: Shouldn't we still open the Help Scout beacon here?

packages/web-app/src/modules/helpscout/models/HelpScout.tsx
```diff
@@ -0,0 +1,27 @@
...
+declare function Beacon(method: 'config', formObject: {}): void
+declare function Beacon(method: 'info'): void
+
+declare global {
+  interface Window {
+    Beacon?: typeof Beacon
+    // https://support.trustpilot.com/hc/en-us/articles/115011421468--Add-a-TrustBox-to-a-single-page-application
+    Trustpilot?: {
+      loadFromElement: (ref: Ref<any>, flag: boolean) => void
+    }
```
seniorquico, line 25: Is this now defined twice? It probably shouldn't be part of the Help Scout types.

packages/web-app/src/modules/machine-views/DontLoseProgressContainer.ts
```diff
@@ -5,7 +5,6 @@ import { DontLoseProgressPage } from './components/DontLoseProgress'
 
 const mapStoreToProps = (store: RootStore): any => ({
   onCloseClicked: () => store.ui.hideModal(true),
-  onSubmitSupportTicket: () => store.zendesk.openSupportTicket(),
```
seniorquico, line 8: Shouldn't we still open the Help Scout beacon here?

packages/web-app/src/modules/machine-views/components/DontLoseProgress.tsx
```diff
@@ -38,14 +39,13 @@ const styles = (theme: SaladTheme) => ({
 
 interface Props extends WithStyles<typeof styles> {
   prepTime: number
-  onSubmitSupportTicket: () => void
```
seniorquico, line 41: Shouldn't we still open the Help Scout beacon here?

packages/web-app/src/modules/start-button/StartButtonUIStore.tsx
```diff
@@ -72,7 +72,7 @@ export class StartButtonUIStore {
 
     const notConnected = () => {
       if (this.supportNeeded) {
-        this.store.zendesk.openSupportTicket()
```
seniorquico, line 75: Shouldn't we still open the Help Scout beacon here?

packages/web-app/src/modules/onboarding-views/OnboardingSpecificAntivirusGuideContainer.tsx
```diff
@@ -1,8 +1,6 @@
 import { connect } from '../../connect'
 import { RootStore } from '../../Store'
-import { getSanitizedHTML } from '../../utils'
```
seniorquico, line 3: It looks like we've removed a lot of references to this `getSanitizedHTML` function. Is it still being used in the app, or should it also be removed?

seniorquico (changes requested):

packages/web-app/src/modules/helpscout/models/HelpScout.tsx
jpark799: it's only defined once. it was original in a model file under zendesk but i removed all those. i'll move this to an onboarding models file because we're using it in the onboarding

packages/web-app/src/modules/machine-views/DontLoseProgressContainer.ts
jpark799: the help scout beacon will still be there but because of the z-index being 2000 on the `DontLoseProgress` page and `FallBackErrorPage` it isn't viewable. given that this is in the legacy app, i don't know if it's worth the effort to rework these. i did add links to the support page though.

packages/web-app/src/modules/error-views/components/FallbackErrorPage.tsx
jpark799: the help scout beacon will still be there but because of the z-index being 2000 on the `DontLoseProgress` page and `FallBackErrorPage` it isn't viewable. given that this is in the legacy app, i don't know if it's worth the effort to rework these. i did add links to the support page though.

packages/web-app/src/modules/error-views/FallbackErrorContainer.ts
jpark799: the help scout beacon will still be there but because of the z-index being 2000 on the `DontLoseProgress` page and `FallBackErrorPage` it isn't viewable. given that this is in the legacy app, i don't know if it's worth the effort to rework these. i did add links to the support page though.

packages/web-app/src/modules/start-button/StartButtonUIStore.tsx
jpark799: I may be mistaken but isn't the `openSupportTicket` function only available for zendesk? 
we will still be able to view the beacon as this will be displayed on the store page.

seniorquico (approved):

PR #951, opened 2022-09-15
[App] Master Branch Discount labels in store

jpark799:
Adds discount labels to our storefront. 

Hero Item:
image
Top Chops: 
image
Reward Details Page:
image
Search Results:
image

### DIFF
--- a/packages/web-app/src/modules/reward-views/components/RewardHeaderBar.tsx
+++ b/packages/web-app/src/modules/reward-views/components/RewardHeaderBar.tsx
@@ -4,6 +4,7 @@ import withStyles, { WithStyles } from 'react-jss'
 import { Button, SmartLink } from '../../../components'
 import { SaladTheme } from '../../../SaladTheme'
 import { Reward } from '../../reward/models'
+import { getPercentOff } from '../../reward/utils'
 import { IconArrowLeft } from './assets'
 
 const styles = (theme: SaladTheme) => ({
@@ -83,6 +84,22 @@ const styles = (theme: SaladTheme) => ({
     color: theme.darkBlue,
     backgroundColor: theme.green,
   },
+  discountLabel: {
+    fontFamily: theme.fontGroteskBook19,
+    color: theme.darkBlue,
+    backgroundColor: theme.cyan,
+    fontSize: 12,
+    letterSpacing: 1,
+    paddingLeft: 5,
+    paddingRight: 10,
+    marginRight: 8,
+    marginTop: 3,
+  },
+  originalPrice: {
+    textDecoration: 'line-through',
+    opacity: 0.5,
+    color: theme.green,
+  },
 })
 
 interface Props extends WithStyles<typeof styles> {
@@ -152,21 +169,31 @@ class _RewardHeaderBar extends Component<Props> {
         {!promoGame && (
           <>
             <div className={classes.priceContainer}>
-              <div
-                className={classnames(classes.priceText, {
-                  [classes.outOfStockPrice]: hasBalance,
-                  [classes.outOfStockPrice]: outOfStock,
-                })}
-              >
-                ${reward ? reward.price.toFixed(2) : '-'}
-              </div>
-
+              {reward && reward.originalPrice && !outOfStock ? (
+                <span className={classes.priceText}>
+                  <span className={classes.originalPrice}>{reward.originalPrice}</span> {reward.price}
+                </span>
+              ) : (
+                <div
+                  className={classnames(classes.priceText, {
+                    [classes.outOfStockPrice]: hasBalance,
+                    [classes.outOfStockPrice]: outOfStock,
+                  })}
+                >
+                  ${reward ? reward.price.toFixed(2) : '-'}
+                </div>
+              )}
+              {reward && reward.originalPrice && !outOfStock && (
+                <div className={classnames(classes.discountLabel)}>
+                  {getPercentOff(reward.originalPrice, reward.price)}{' '}
+                </div>
+              )}
               {outOfStock && (
                 <div className={classnames(classes.priceText, classes.stockLabel, classes.outOfStockLabel)}>
                   Out of Stock
                 </div>
               )}
-              {lowQuanity && (
+              {lowQuanity && !reward?.originalPrice && (
                 <div className={classnames(classes.priceText, classes.stockLabel, classes.lowQuanityLabel)}>
                   {`${reward?.quantity} Remaining`}
                 </div>
--- a/packages/web-app/src/modules/reward-views/components/RewardHeaderBar.tsx
+++ b/packages/web-app/src/modules/reward-views/components/RewardHeaderBar.tsx
@@ -170,9 +170,9 @@ class _RewardHeaderBar extends Component<Props> {
           <>
             <div className={classes.priceContainer}>
               {reward && reward.originalPrice && !outOfStock ? (
-                <span className={classes.priceText}>
+                <div className={classes.priceText}>
                   <span className={classes.originalPrice}>{reward.originalPrice}</span> {reward.price}
-                </span>
+                </div>
               ) : (
                 <div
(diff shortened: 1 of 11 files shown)


packages/web-app/src/modules/reward-views/components/RewardHeaderBar.tsx
```diff
@@ -83,6 +84,22 @@ const styles = (theme: SaladTheme) => ({
...
     backgroundColor: theme.green,
   },
+  discountLabel: {
+    fontFamily: theme.fontGroteskBook19,
+    color: theme.darkBlue,
+    backgroundColor: theme.cyan,
+    fontSize: 12,
+    letterSpacing: 1,
+    paddingLeft: 5,
+    paddingRight: 10,
```
scottcinatl, line 94: Is it by design to have more padding to the right of the percentage than to the left? It seems odd that the label doesn't have the content centered in the box.

packages/web-app/src/modules/reward-views/components/RewardHeaderBar.tsx
```diff
@@ -152,21 +169,31 @@ class _RewardHeaderBar extends Component<Props> {
...
+                  {getPercentOff(reward.originalPrice, reward.price)}{' '}
+                </div>
+              )}
               {outOfStock && (
                 <div className={classnames(classes.priceText, classes.stockLabel, classes.outOfStockLabel)}>
                   Out of Stock
                 </div>
               )}
-              {lowQuanity && (
+              {lowQuanity && !reward?.originalPrice && (
```
scottcinatl, line 196: I could understand on the `RewardItem` shown in the storefront that we would want to optimize space (curious about whether low stock is more important information to the user rather than being explicit with showing how much of a discount the user is getting), but for the header it seems like there is ample space to have both labels showing. Not sure if that is desired?

packages/web-app/src/modules/reward-views/components/RewardHeaderBar.tsx
```diff
@@ -152,21 +169,31 @@ class _RewardHeaderBar extends Component<Props> {
...
-                className={classnames(classes.priceText, {
-                  [classes.outOfStockPrice]: hasBalance,
-                  [classes.outOfStockPrice]: outOfStock,
-                })}
-              >
-                ${reward ? reward.price.toFixed(2) : '-'}
-              </div>
-
+              {reward && reward.originalPrice && !outOfStock ? (
+                <span className={classes.priceText}>
```
scottcinatl, line 173: nit: nested spans are valid HTML, but this breaks away from the pattern that is used in the other labels where the `div` handles the style of the label and the `span` handles the styles for what is inside the label.

packages/web-app/src/modules/reward-views/components/RewardItem.tsx
```diff
@@ -120,15 +137,26 @@ class _RewardItem extends Component<Props> {
...
-            <div className={classnames(classes.priceText, { [classes.outOfStockPrice]: outOfStock })}>
-              {reward ? reward?.price ? `$${reward?.price.toFixed(2)}` : 'FREE' : <Skeleton width={100} />}
-            </div>
+            {reward && reward?.originalPrice && !outOfStock && (
+              <div className={classnames(classes.discountLabel)}>
+                {getPercentOff(reward.originalPrice, reward.price)}{' '}
+              </div>
+            )}
+            {reward && reward.originalPrice && !outOfStock ? (
+              <span className={classes.priceText}>
```
scottcinatl, line 146: nit: nested spans that break from existing pattern

packages/web-app/src/modules/reward/models/Reward.ts
```diff
@@ -28,6 +28,9 @@ export interface Reward {
   /** Purchase price */
   price: number
 
+  /** MSRP of the reward */
+  originalPrice?: number
```
scottcinatl, line 32: I see that we had to make this change in two different models, but to me there doesn't seem to be a difference in the `Reward` model and the `RewardResource` model, if they are the same/reference the same thing we should only have one of them.

packages/web-app/src/modules/reward/models/SearchResult.ts
```diff
@@ -15,12 +15,14 @@ export class SearchResult {
...
     public readonly description?: string,
     public readonly quantity?: number,
+    public readonly originalPrice?: number,
   ) {}
 
   public static parseSearchResult = (result: any): SearchResult => {
     const id = result['id'].raw
     const name = result['name'].raw
     const price = parseFloat(result['price'].raw)
+    let originalPrice: number | undefined = parseFloat(result['original_price']?.raw)
```
scottcinatl, line 25: So the types here are incorrect, because we would be assigning `NaN` to original price in which we are typing as either `number | undefined`. I know typescript doesn't have a type for `NaN` and you are catching that case down below. But I think this could be cleaner if before applying the `parseFloat` method that we check if the value is defined first, if not, set it to `undefined`. If it is, `parseFloat` like there is no tomorrow :)

packages/web-app/src/modules/reward-views/components/RewardItem.tsx
```diff
@@ -120,15 +137,26 @@ class _RewardItem extends Component<Props> {
         <div className={classes.textContainer}>
           <div className={classes.nameText}>{reward ? reward.name : <Skeleton />}</div>
           <div className={classes.subTextContainer}>
-            <div className={classnames(classes.priceText, { [classes.outOfStockPrice]: outOfStock })}>
-              {reward ? reward?.price ? `$${reward?.price.toFixed(2)}` : 'FREE' : <Skeleton width={100} />}
-            </div>
+            {reward && reward?.originalPrice && !outOfStock && (
```
scottcinatl, line 140: Since `reward` is already been checked for we don't have to have utilize optional chaining with `reward?.originalPrice` , it could just be `reward.originalPrice`

packages/web-app/src/modules/storefront-views/components/StorefrontRewardItem.tsx
```diff
@@ -123,15 +144,26 @@ class _StorefrontRewardItem extends Component<Props> {
         <div className={classes.textContainer}>
           <div className={classes.nameText}>{name ? name : <Skeleton />}</div>
           <div className={classes.subTextContainer}>
-            <div className={classnames(classes.priceText, { [classes.outOfStockPrice]: outOfStock })}>
-              {price ? price : <Skeleton width={100} />}
-            </div>
+            {rewardData && rewardData?.originalPrice && !outOfStock && (
```
scottcinatl, line 147: unnecessary optional chaining

scottcinatl (changes requested):

packages/web-app/src/modules/reward-views/components/RewardHeaderBar.tsx
jpark799: that's how the figma specs were. i think the idea was to have the digits in the center not everything including the `-`.

packages/web-app/src/modules/reward-views/components/RewardHeaderBar.tsx
jpark799: yeah that's a good point. I can double check with Armin

packages/web-app/src/modules/reward/models/Reward.ts
jpark799: it looks like there's a couple of difference between the two like `Reward` having a `videos` and checkoutTerms` field. I agree that these probably couple be consolidated but i think that's outside the scope of this PR.

packages/web-app/src/modules/reward-views/components/RewardHeaderBar.tsx
jpark799: checked with armin and he said it felt like there was too much clutter with two tags. i also agree 
image

PR #959, opened 2022-11-08
[App - Master Branch] Add must update notifications banner 

Maks19:
**Description:**
Add must update notifications banner to legacy app

image

**Issues:** https://www.notion.so/saladtech/Create-a-banner-with-text-347b4af2021f491ea656aacbc908ee8d ,
https://www.notion.so/saladtech/Implement-a-banner-in-every-section-that-can-support-showing-it-d866700efe9f46309b598d75ca7ef789


jpark799 (approved):

jpark799 (changes requested):
sorry, after looking more closely at the design docs, it looks like we have a little too much padding on the bottom of the banner. can you please update this to reflect what is in the figma? or if ivan/simina says it's not a big deal, i can just approve. just want to make sure we're sticking as close as we can to the designs product is giving us. thank you!


image

image

Maks19:
> sorry, after looking more closely at the design docs, it looks like we have a little too much padding on the bottom of the banner. can you please update this to reflect what is in the figma? or if ivan/simina says it's not a big deal, i can just approve. just want to make sure we're sticking as close as we can to the designs product is giving us. thank you!
> 
> image
> 
> image

You're right
I fixed it

jpark799 (approved):

Maks19:
Changes to alpha branch should be enough for the legacy app
Closed after a conversation with Jason

PR #971, opened 2022-12-06
Reword redemption xp error

FemboyNova:
Because of the numerous support messages in the discord about how much XP and time on Salad is actually required to redeem a reward, I believe the error message should be rewritten.

I changed it to the message I believe it best fits this, but feel free to change it however you see fit as long as it fits the theme of knowing how much time is required to redeem a reward, and I believe many others will be pleased.

Another way it could be worded is, "This Salad account is too new. You need 1440 XP to redeem rewards, keep chopping and try again later." 

Also note that I havent been able to check how this will fit in the box, so please do so before doing the merge.

### DIFF
--- a/packages/web-app/src/modules/reward/RewardStore.ts
+++ b/packages/web-app/src/modules/reward/RewardStore.ts
@@ -347,7 +347,7 @@ export class RewardStore {
                 notification = {
                   category: NotificationMessageCategory.Error,
                   title: 'Redemption Error',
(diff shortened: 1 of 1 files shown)


ivanatsalad:
The specific amount of XP required is variable and the requirement itself has security implications. We will not be disclosing it specifically for now.

FemboyNova:
> 

Okay thanks for letting me know, i will close this pr now

PR #974, opened 2022-12-19
[App] Add users last seen t&c's version update to login flow

rners01:
project - https://www.notion.so/saladtech/Terms-Conditions-Update-Flow-565e181573ff4068bc7aa5881a798534
task - https://www.notion.so/saladtech/App-Add-login-register-update-users-last-seen-terms-e90af1576e874adb828e691200beee6a

### DIFF
--- a/packages/web-app/src/Store.ts
+++ b/packages/web-app/src/Store.ts
@@ -99,7 +99,6 @@ export class RootStore {
   constructor(axios: AxiosInstance, private readonly featureManager: FeatureManager) {
     this.routing = new RouterStore()
     this.auth = new AuthStore(axios, this.routing)
-    this.termsAndConditions = new TermsAndConditionsStore(axios, this)
     this.notifications = new NotificationStore(this)
     this.xp = new ExperienceStore(axios)
     this.native = new NativeStore(this)
@@ -108,6 +107,7 @@ export class RootStore {
 
     this.machine = new MachineStore(this, axios, featureManager)
     this.profile = new ProfileStore(this, axios)
+    this.termsAndConditions = new TermsAndConditionsStore(axios, this.profile)
     this.saladCard = new SaladCardStore(this, axios)
     this.rewards = new RewardStore(this, axios, this.profile, this.saladCard)
     this.analytics = new AnalyticsStore(this)
@@ -192,6 +192,10 @@ export class RootStore {
         this.helpScout.login(profile.username, profile.email),
         Promise.race([this.saladBowl.login(), delay(10000)]),
       ])
+
+      if(profile.pendingTermsVersion) {
+        yield this.termsAndConditions.submitTermsAndConditions()
+      }
 
       this.finishInitialLoading()
       this.onboarding.showOnboardingIfNeeded()
--- a/packages/web-app/src/modules/terms-and-conditions/TermsAndConditionsStore.ts
+++ b/packages/web-app/src/modules/terms-and-conditions/TermsAndConditionsStore.ts
@@ -1,6 +1,6 @@
 import { AxiosInstance, AxiosResponse } from 'axios'
 import { action, flow, observable } from 'mobx'
-import { RootStore } from '../../Store'
+import { ProfileStore } from '../profile'
 import { Profile } from '../profile/models'
 
 export class TermsAndConditionsStore {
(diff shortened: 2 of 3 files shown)


packages/web-app/src/modules/terms-and-conditions/TermsAndConditionsStore.ts
```diff
@@ -14,20 +14,26 @@ export class TermsAndConditionsStore {
...
+    private readonly profile: ProfileStore,
     ) {}
 
   @action.bound
   public submitTermsAndConditions = flow(function* (this: TermsAndConditionsStore) {
     try {
+      const pendingTermsVersion = this.profile.currentProfile?.pendingTermsVersion
+
+      if(!pendingTermsVersion) {
+        return
```
Maks19, line 26: Could we log what went wrong here?
As I understand `submitTermsAndConditions` will be triggered after button click so if `pendingTermsVersion` is false user won't see anything and we won't know what happened

packages/web-app/src/Store.ts
```diff
@@ -189,6 +189,7 @@ export class RootStore {
         this.refresh.refreshData(),
         this.profile.loadPayPalId(),
         this.saladCard.loadSaladCard(),
+        this.termsAndConditions.submitTermsAndConditions(),
```
scottcinatl, line 192: What is the expected experience on an error here? It seems that in the `submitTermsAndConditions` call that we are not throwing any error if one occurs. Is the user supposed to be able to continue if they aren't able to accept the new terms and conditions?

packages/web-app/src/modules/terms-and-conditions/TermsAndConditionsStore.ts
rners01: no, this is for other purposes(Im reusing this logic for login/register flow, there could be no ```pendingTermsVersion``` so I don't want to send request)
if ```pendingTermsVersion``` is ```null```, then user wont even see this whole block with update T&Cs

packages/web-app/src/Store.ts
rners01: good question, with existing logic
if user will not be able to update T&Cs he will just keep seeing notification in the widget to update them and update T&Cs block in the account page in web-app

rners01: I'll double check that with Ivan, to be sure that such behaviour satisfies is acceptable

rners01: yeah, looks like Ivan will be back after holidays
What do you think then @scottcinatl ?

rners01: checked with Ivan, he seems ok with existing solution
image

packages/web-app/src/modules/terms-and-conditions/TermsAndConditionsStore.ts
rners01: @Maks19 updated the solution to drop additional check

scottcinatl (changes requested):

packages/web-app/src/Store.ts
```diff
@@ -192,6 +192,9 @@ export class RootStore {
         this.helpScout.login(profile.username, profile.email),
         Promise.race([this.saladBowl.login(), delay(10000)]),
       ])
+      if(profile.pendingTermsVersion) {
```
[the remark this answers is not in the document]
rners01: its all possible, but I wanted to keep `submitTermsAndConditions` clean and simple, as we'r reusing it in couple places

scottcinatl (dismissed):

packages/web-app/src/Store.ts
scottcinatl, line 195: I noticed this wasn't automatically formatted by prettier. There should be a space here, when you save. So make sure we are opening the project in the right folder so we get the correct vscode tooling. For instance, I think this project has to be opened in the `salad-applications`  folder. But when jumping around to each project let's just make sure we are opening it from the correct folder.

rners01: Its not working, where do you see such config, I can't find it

rners01: but anyway fixed,
we should add some eslint config file for it

PR #975, opened 2022-12-21
[App] Add lifetimeXP to HelpScout beacon API

rners01:
project - https://www.notion.so/saladtech/Helpscout-Beacon-NPS-Targeting-1dd0a27e51714f6495599195886b252f
task - https://www.notion.so/saladtech/App-Add-lifetimeXP-to-Beacon-API-83ac94197ee0485dbfc4e7cf0baa68a0

### DIFF
--- a/packages/web-app/src/Store.ts
+++ b/packages/web-app/src/Store.ts
@@ -186,7 +186,11 @@ export class RootStore {
         this.refresh.refreshData(),
         this.profile.loadPayPalId(),
         this.saladCard.loadSaladCard(),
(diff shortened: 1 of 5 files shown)


scottcinatl (approved):

scottcinatl:
@rners01 We have two versions of the app that we maintain, So I would think this change would also need to be made on the [alpha branch](https://github.com/SaladTechnologies/salad-applications/tree/alpha). 

Anytime a change is made in this repo, that will have to be a consideration. Sometimes changes are made for both, other times not. So something to bring up with Ivan for any of these salad application tasks.

PR #976, opened 2023-01-04
[Web-App] Seasons section - removed from account/summary sidebar

vitto-moz:
https://www.notion.so/saladtech/Hide-Seasons-section-in-the-user-profile-89816fe52f7a484cab666cb625e461c8

Seasons section - removed from `account/summary` sidebar

### DIFF
--- a/packages/web-app/src/modules/settings-views/SettingsContainer.ts
+++ b/packages/web-app/src/modules/settings-views/SettingsContainer.ts
@@ -5,21 +5,18 @@ import { RootStore } from '../../Store'
 import { AccountContainer } from '../account-views/account-views'
 import { ReferralSettingsContainer } from '../account-views/referral-views'
 import { BonusPageContainer } from '../bonus-views'
-import { CurrentSeasonPageContainer } from '../seasons-views'
 import { DesktopSettingsContainer } from './desktop-settings-views'
 
 const mapStoreToProps = (store: RootStore): any => {
   const menuItems: MenuItem[] = [
     { url: '/account/summary', text: 'Account', component: AccountContainer },
     { url: '/account/referrals', text: 'Referrals', component: ReferralSettingsContainer },
-    { url: '/account/seasons', text: 'Seasons', component: CurrentSeasonPageContainer, divider: true },
     { url: '/account/bonuses', text: 'Bonuses', component: BonusPageContainer },
   ]
 
   const nativeMenuItems: MenuItem[] = [
     { url: '/account/summary', text: 'Account', component: AccountContainer },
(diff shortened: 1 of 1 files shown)


packages/web-app/src/modules/settings-views/SettingsContainer.ts
```diff
@@ -5,21 +5,21 @@ import { RootStore } from '../../Store'
 import { AccountContainer } from '../account-views/account-views'
 import { ReferralSettingsContainer } from '../account-views/referral-views'
 import { BonusPageContainer } from '../bonus-views'
-import { CurrentSeasonPageContainer } from '../seasons-views'
+// import { CurrentSeasonPageContainer } from '../seasons-views'
```
scottcinatl, line 8: These should probably just be deleted rather than commented out.

scottcinatl (changes requested):

packages/web-app/src/modules/settings-views/SettingsContainer.ts
scottcinatl: There is also an alpha branch version of this app, I would think the same changes need to be made in that branch. 

Should this change also include an updating of the navigation bar with the removal of the links to seasons as well? So it is removed in one PR rather than two? Or is Ivan fine with this removal not all happening in the same update?

scottcinatl:
@vitto-moz We have two versions of this app. Sometimes changes need to be to both this branch and the [alpha branch/legacy](https://github.com/SaladTechnologies/salad-applications/tree/alpha). I would check with Ivan to see if this is the case for this change.

packages/web-app/src/modules/settings-views/SettingsContainer.ts
vitto-moz: > These should probably just be deleted rather than commented out.

@scottcinatl 
Just commented, as we temporary hide this functionality.
We can add some comments, or do you still think we have to completely remove it?

vitto-moz: @scottcinatl nevermind about my previous comment 
I've found this comment of @seniorquico, for the related PR 
https://github.com/SaladTechnologies/salad-platform/pull/2507#discussion_r1064888607

scottcinatl (approved):

packages/web-app/src/modules/settings-views/SettingsContainer.ts
scottcinatl: I approved. But still had a question about whether this PR should also include the garden component package update which also removes the links stuff from the navigation.

Maks19 (approved):

PR #978, opened 2023-01-10
Hide seasons section in the user profile

vitto-moz:
https://www.notion.so/saladtech/Hide-Seasons-section-in-the-user-profile-89816fe52f7a484cab666cb625e461c8


vitto-moz:
Wrong source of branching

PR #987, opened 2023-03-10
Remove most of the obsolete desktop app references

seniorquico:
This removes most of the obsolete desktop app references from the web app. This shifts the web app into most of what it is today: the way for Chefs to sign-up or sign-in, view their Salad Balance, and redeem rewards.


packages/web-app/src/modules/home-views/NavigationBarContainer.tsx
```diff
@@ -42,7 +39,7 @@ const mapStoreToProps = (store: RootStore): any => {
         isLoading={store.bonuses.pendingBonuses?.has(bonus.id)}
       />
     ) : undefined,
-    isNative: isNative,
+    isNative: false,
```
Maks19, line 42: Could we just remove `isNative`?

packages/web-app/package.json
```diff
@@ -55,122 +46,144 @@
...
-    "@fortawesome/free-regular-svg-icons": "6.1.2",
-    "@fortawesome/free-solid-svg-icons": "6.1.2",
+    "@emotion/react": "11.10.6",
+    "@emotion/styled": "11.10.6",
+    "@fortawesome/fontawesome-svg-core": "6.3.0",
+    "@fortawesome/free-regular-svg-icons": "6.3.0",
+    "@fortawesome/free-solid-svg-icons": "6.3.0",
     "@fortawesome/react-fontawesome": "0.2.0",
-    "@saladtechnologies/garden-components": "1.0.86",
+    "@saladtechnologies/garden-components": "1.0.89",
```
Maks19, line 82: Could you stay with `1.0.86` version?
I found a little bug for web-app in garden-components latest library

image

packages/web-app/src/modules/earn-views/pages/EarningsSummaryPage.stories.tsx
```diff
@@ -1,15 +1,13 @@
-import {
-  HardwareCardProps,
-  HardwareDetailSingleColumn,
-  HardwareDetailTwoColumns,
-} from '@saladtechnologies/garden-components'
+import type { HardwareCardProps } from '@saladtechnologies/garden-components'
```
Maks19, line 1: Little thing
I think `isNative` can be also removed from this file

packages/web-app/src/modules/earn-views/components/MiningInformation.stories.tsx
```diff
@@ -1,4 +1,4 @@
-import { Meta } from '@storybook/react'
+import type { Meta } from '@storybook/react'
```
Maks19, line 1: Little thing
I think `isNative` can be also removed from this file

packages/web-app/src/modules/home-views/NavigationBarContainer.tsx
seniorquico: `isNative` is a required prop on the `NavigationBar` component, which is part of the Garden. I cannot remove this prop here/now, but- Yes, it can be removed from the Garden and then, eventually, removed from the web app.

packages/web-app/package.json
seniorquico: I've reverted it in this PR. However, this is problematic- There are garden changes happening on other projects right now. It sounds like this is a defect that will need to be prioritized!

PR #1007, opened 2023-07-26
Added DevOps access PAT support

shoneefd:
I have... limited confidence that this will work. Ideally, the addition of this token in the environment of the initialize script should make the read work properly. But it's possible I may need to do the environment injection during NPM login instead.


sunuk-salad (dismissed):

seniorquico:
The idea of this should work... But aren't there built-in tasks that login to npm? We use them in a couple other pipelines.

PR #1019, opened 2023-09-05
[Web-app] Add novu hmac signature & change novu notifications stream logic

rners01:
**Description:** Change novu notifications stream logic and Add novu hmac signature - https://docs.novu.co/notification-center/react/react-components/#enabling-hmac-encryption
**Notion Task:** https://www.notion.so/saladtech/App-Add-novu-hmac-signature-8194f4321f6c44c390d27cf7a66d840a
**Notion Project:** https://www.notion.so/saladtech/Frontend-Add-push-notifications-to-the-web-app-3a60520ff8d4485d85965b12fa1c2e84


packages/web-app/src/modules/profile/ProfileStore.tsx
```diff
@@ -77,6 +80,18 @@ export class ProfileStore {
...
+  @action.bound
+  loadNovuSignature = flow(function* (this: ProfileStore) {
+    try {
+      if (this.novuSignature) {
+        return
+      }
+
+      let novuSignature = yield this.axios.post('/api/v2/novu-signatures')
+      this.novuSignature = novuSignature.data.signature
+    } catch (err) {}
```
Maks19, line 92: Should there be some logging error logic here?

rners01: we don't have any logging example

rners01: so the worst case scenario here without this signature -> initialization fails -> our alerts block would be empty

PR #1030, opened 2023-09-13
[WebApp] new endpoints used for login and logout / initial check if user is au…

vitto-moz:
https://www.notion.so/saladtech/Implement-Identity-Core-authentication-functionality-but-preserve-SuperTokens-ab4fe9c41cca48ca9f2f463c5fdeeb7b?pvs=4

### DIFF
--- a/packages/web-app/src/modules/auth/AuthStore.tsx
+++ b/packages/web-app/src/modules/auth/AuthStore.tsx
@@ -73,9 +73,14 @@ export class AuthStore {
 
     SuperTokens.doesSessionExist()
       .then((result) => {
-        runInAction(() => {
-          this.isAuthenticated = result
-        })
+        if (result) {
+          runInAction(() => (this.isAuthenticated = result))
+        } else {
+          this.axios
+            .get('/api/v1/profile')
+            .then(() => runInAction(() => (this.isAuthenticated = true)))
+            .catch(() => runInAction(() => (this.isAuthenticated = false)))
+        }
       })
       .catch(() => {})
   }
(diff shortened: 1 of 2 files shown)


packages/web-app/src/modules/auth/AuthStore.tsx
```diff
@@ -73,9 +73,14 @@ export class AuthStore {
...
-        runInAction(() => {
-          this.isAuthenticated = result
-        })
+        if (result) {
+          runInAction(() => (this.isAuthenticated = result))
+        } else {
+          this.axios
+            .get('/api/v1/profile')
+            .then(() => runInAction(() => (this.isAuthenticated = true)))
+            .catch(() => runInAction(() => (this.isAuthenticated = false)))
```
rners01, line 82: Does in `then` will come only 2xx status?
And that means that we will get profile data?

packages/web-app/src/modules/auth/AuthStore.tsx
```diff
@@ -149,7 +156,7 @@ export class AuthStore {
 
       this.isSubmitting = true
 
-      yield this.axios.post('/api/v2/authentication-sessions', request)
+      yield this.axios.post('/api/v2/user-accounts/login', request)
```
rners01, line 159: I remember Kyle said that we don't need to change the URLs
because when we will be ready to release -> backend will substitute existing API methods with the new logic

packages/web-app/src/modules/auth/AuthStore.tsx
vitto-moz: [yes, axios considers any other codes than 2XX as an error ](https://axios-http.com/docs/handling_errors)

packages/web-app/src/modules/auth/AuthStore.tsx
vitto-moz: yeah, correct, but we want a test as much closet to the real environment as possible

rners01: so you will release it only to test server?

PR #1039, opened 2023-09-20
[Web-App] Remove supertokens

Maks19:
**Description:** Remove supertokens

**Notion Task:** https://www.notion.so/saladtech/Web-App-remove-SuperTokens-2992c392035c405194b6e56cb527ba19

### DIFF
--- a/packages/web-app/src/axiosFactory.tsx
+++ b/packages/web-app/src/axiosFactory.tsx
@@ -2,7 +2,6 @@ import type { AxiosError, AxiosInstance, AxiosResponse } from 'axios'
 import axios from 'axios'
 import axiosRetry, { exponentialDelay } from 'axios-retry'
 import isRetryAllowed from 'is-retry-allowed'
-import SuperTokens from 'supertokens-website'
 import { config } from './config'
 import { getStore } from './Store'
 
@@ -32,10 +31,10 @@ const shouldRetryDownload = (error: any): boolean => {
 
 export const createClient = (): AxiosInstance => {
   let httpClient = axios.create({
+    withCredentials: true,
     baseURL: config.apiBaseUrl,
   })
 
-  SuperTokens.addAxiosInterceptors(httpClient)
   httpClient.interceptors.response.use(
     (response) => {
       // Any status code that lie within the range of 2xx cause this function to trigger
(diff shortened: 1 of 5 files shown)


packages/web-app/src/axiosFactory.tsx
```diff
@@ -32,10 +31,10 @@ const shouldRetryDownload = (error: any): boolean => {
 
 export const createClient = (): AxiosInstance => {
   let httpClient = axios.create({
+    withCredentials: true,
```
vitto-moz, line 34: is this for third-party requests? 
do we use axios to request third-party services?

Maks19: It was done for requests should be made using credentials such as cookies, authorization headers or TLS client certificates
Without this parameter, our axios requests won't include `auth` token

seniorquico: Is this httpClient used to make requests to any API other than ours? If yes, then ideally this value is set true only for those requests.

packages/web-app/src/axiosFactory.tsx
```diff
@@ -45,12 +44,8 @@ export const createClient = (): AxiosInstance => {
...
       try {
         if (error.response.status === 401 && error.config.baseURL === config.apiBaseUrl) {
-          SuperTokens.doesSessionExist().then((result) => {
-            if (!result) {
-              const store = getStore()
-              store.auth.setIsAuthenticated(false)
-            }
-          })
+          const store = getStore()
+          store.auth.setIsAuthenticated(false)
```
seniorquico, line 48: Is the httpClient this interceptor attached to used to make requests to any API other than ours? If yes, then a failing request to another API will cause the app to logout. The removed SuperTokens session check first verified that it was actually expired before causing the app to logout.

seniorquico (changes requested):

packages/web-app/src/axiosFactory.tsx
Maks19: We're not making any third-party requests with an axios and the `httpClient` is axios instance so it's not making requests to any API other than ours

packages/web-app/src/axiosFactory.tsx
Maks19: We're not making any third-party requests with an axios so this interceptor is not attached to requests to any API other than ours

seniorquico (approved):

vitto-moz (approved):

PR #1066, opened 2023-10-26
[WebApp] Achievements: feature flag condition - added

vitto-moz:
https://www.notion.so/saladtech/WebApp-Add-Unleash-flag-91cd8a956aa54cbf9ad50dba9c048ac2?pvs=4

### DIFF
--- a/packages/web-app/src/components/SettingsPage.tsx
+++ b/packages/web-app/src/components/SettingsPage.tsx
@@ -44,8 +44,8 @@ const _Settings = ({ appBuild, classes, menuButtons, onClose }: Props) => {
   }
 
   const handleCloseKeyPress = useCallback(
-    (e: KeyboardEvent) => {
-      if (e.key === 'Escape') {
+    (event: KeyboardEvent) => {
+      if (event.key === 'Escape') {
         onClose?.()
       }
     },
--- a/packages/web-app/src/components/SettingsPage.tsx
+++ b/packages/web-app/src/components/SettingsPage.tsx
@@ -1,10 +1,14 @@
 import classnames from 'classnames'
-import type { ComponentType, ReactNode } from 'react'
-import { Component } from 'react'
+import { useCallback, useEffect, type ComponentType } from 'react'
 import type { WithStyles } from 'react-jss'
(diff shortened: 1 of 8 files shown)


packages/web-app/src/components/SettingsPage.tsx
```diff
@@ -28,64 +32,74 @@ export interface MenuButton {
...
-    const { onClose } = this.props
+  const handleCloseClicked = () => {
     onClose?.()
   }
 
-  private handleCloseKeyPress = (e: any) => {
-    if (e.key === 'Escape') {
-      const { onClose } = this.props
-      onClose?.()
+  const handleCloseKeyPress = useCallback(
```
rners01, line 46: why do we need `useCallback` here?

packages/web-app/src/components/SettingsPage.tsx
```diff
@@ -28,64 +32,74 @@ export interface MenuButton {
...
     }
-  }
+  }, [handleCloseKeyPress])
+
+  const menuItems: MenuItem[] = [
+    { url: '/account/summary', text: 'Account', component: AccountContainer },
+    { url: '/account/referrals', text: 'Referrals', component: ReferralSettingsContainer },
+    { url: '/account/bonuses', text: 'Bonuses', component: BonusPageContainer },
+    isAchievementsFeatureFlagEnabled && {
+      url: '/account/achievements',
```
rners01, line 67: what would we show if user is not logged in?

packages/web-app/src/components/SettingsPage.tsx
vitto-moz: this function is used as a dependency for `useEffect`
```
 useEffect(() => {
    document.addEventListener('keyup', handleCloseKeyPress)
    return () => {
      document.removeEventListener('keyup', handleCloseKeyPress)
    }
  }, [handleCloseKeyPress])

packages/web-app/src/components/SettingsPage.tsx
vitto-moz: image

Maks19 (approved):

PR #1098, opened 2023-12-14
[WebApp] Top space for page content 

vitto-moz:
Layout issue - fixed
image

### DIFF
--- a/packages/web-app/src/App.tsx
+++ b/packages/web-app/src/App.tsx
@@ -104,25 +104,15 @@ const searchConfig = {
 
 interface AppProps extends WithStyles<typeof styles> {
   isAuthenticated: boolean
+  withInstallReminder: boolean
   novuSignature: string
-  saladBowlFirstLoginAt: string
-  isInstallReminderClosed: boolean
   history: History
 }
 
-export const _App = ({
-  classes,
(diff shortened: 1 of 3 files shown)


packages/web-app/src/App.tsx
```diff
@@ -125,7 +133,12 @@ export const _App = ({ classes, history, isAuthenticated, novuSignature }: AppPr
         <div className={classes.mainWindow}>
           <NavigationBarContainer />
           <div className={classes.container}>
-            <div className={classes.content}>
+            <div
+              className={classNames(
+                classes.content,
+                isNewChefDownloadFeatureFlagEnabled && withInstallReminder && classes.withBanner,
```
Maks19, line 139: Let's create variable that will have value - `isNewChefDownloadFeatureFlagEnabled && withInstallReminder` and use it here for better readiness

vitto-moz: I found the additional variable makes complexity, as I didn't found the convenient name for it. 
Do you have a suggestion for variable name?

Maks19: What about isInstallReminderShown?

Maks19 (approved):

PR #1105, opened 2023-12-18
[WebApp] Connect error page

Maks19:
**Description:** Connect error page

**Issue:** https://www.notion.so/saladtech/Replace-profile-error-with-new-Error-page-bbe2f4cb45d040bd9cb8f51e8cf11030


packages/web-app/src/Store.tsx
```diff
@@ -138,6 +138,7 @@ export class RootStore {
       var profile: Profile = yield this.profile.loadProfile()
       if (!profile) {
         this.auth.logout()
+        this.finishInitialLoading()
```
rners01, line 141: what i this for?

packages/web-app/src/modules/profile/ProfileStore.tsx
```diff
@@ -112,9 +115,10 @@ export class ProfileStore {
     try {
       let profile = yield this.axios.get('/api/v1/profile')
       this.currentProfile = profile.data
+      this.isErrorCaught = false
     } catch (err) {
       this.currentProfile = undefined
-      this.store.routing.replace('/profile-error')
+      this.isErrorCaught = true
```
vitto-moz, line 121: Can we stay with redirect and avoid introducing the new field?

packages/web-app/src/Store.tsx
Maks19: Removed

packages/web-app/src/modules/profile/ProfileStore.tsx
Maks19: We can't stay with redirect because removing redirect was main purpose of the project

packages/web-app/src/modules/error-boundary/ErrorBoundaryStore.tsx
```diff
@@ -0,0 +1,35 @@
+import { action, observable } from 'mobx'
+import type { UseErrorBoundaryApi } from 'react-error-boundary'
+
+export class ErrorBoundaryStore {
+  @observable
+  private showErrorBoundaryQueue: (() => void)[] = []
```
vitto-moz, line 6:
```suggestion
  private pendingShowErrorBoundaryActions: (() => void)[] = []
```

vitto-moz: the variable that starts from `show.. ` looks like a method to me

packages/web-app/src/modules/error-boundary/ErrorBoundaryStore.tsx
```diff
@@ -0,0 +1,35 @@
...
+    this.errorBoundary = errorBoundary
+    // Trigger any queued showErrorBoundary actions
+    this.showErrorBoundaryQueue.forEach((queuedAction) => {
+      queuedAction()
+    })
+    this.showErrorBoundaryQueue = []
+  }
+
+  @action
+  public showErrorBoundary = (errorCaughtMessage: Error) => {
```
vitto-moz, line 22: What do you think about using the same naming as library does

```suggestion
  public showBoundary = (errorCaughtMessage: Error) => {
```

vitto-moz (dismissed):

packages/web-app/src/modules/error-boundary/ErrorBoundaryStore.tsx
Maks19: Agree
Changed

packages/web-app/src/modules/error-boundary/ErrorBoundaryStore.tsx
Maks19: I would prefer showErrorBoundary to know exactly what Boundary should be shown

vitto-moz (approved):

PR #1107, opened 2023-12-21
[WebApp] Redirect from earn to earn summary for authenticated user

rners01:
https://www.notion.so/saladtech/Frontend-Kitchen-Website-Navigation-1b44b7c416e74f4db46cade6c21a6456

### DIFF
--- a/packages/web-app/src/Routes.tsx
+++ b/packages/web-app/src/Routes.tsx
@@ -11,7 +11,11 @@ import { SettingsContainer } from './modules/settings-views'
 import { StorefrontHomePage } from './modules/storefront-views/pages/StorefrontHomePage'
 import { VaultListContainer } from './modules/vault-views'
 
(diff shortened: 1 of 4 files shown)


packages/web-app/src/Routes.tsx
```diff
@@ -38,6 +42,7 @@ const _Routes = ({ location }: RouteComponentProps) => {
       <Redirect exact from="/account" to="/account/summary" />
       <Route path="/account" component={SettingsContainer} />
 
+      {isAuthenticated ? <Redirect exact from="/earn" to="/earn/summary" /> : null}
```
Maks19, line 45: Is it okay to have this route and `<Route path="/earn" component={EarnInfoPage} />`

Maks19, line 45: Can we handle `isAuthenticated` in component?

PR #1118, opened 2024-01-17
[WebApp] Fix search sort value revert issue

rners01:
https://www.notion.so/saladtech/WebApp-Search-sort-value-label-revert-issue-073712e0a784486fbdaddbcec51d336d

### DIFF
--- a/packages/web-app/src/components/Dropdown.tsx
+++ b/packages/web-app/src/components/Dropdown.tsx
@@ -1,7 +1,6 @@
 import type { FC } from 'react'
 import type { WithStyles } from 'react-jss'
 import withStyles from 'react-jss'
-import type { OptionsOrGroups } from 'react-select'
 import Select from 'react-select'
 import type { SaladTheme } from '../SaladTheme'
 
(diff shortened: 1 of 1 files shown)


packages/web-app/src/components/Dropdown.tsx
```diff
@@ -16,45 +15,52 @@ const styles = (theme: SaladTheme) => ({
 
 interface Props extends WithStyles<typeof styles> {
   options?: OptionsOrGroups<any, any>
```
vitto-moz, line 17: It's unclear what the option type should be, could we use any explicit type?

packages/web-app/src/components/Dropdown.tsx
```diff
@@ -16,45 +15,52 @@ const styles = (theme: SaladTheme) => ({
...
-      />
-    )
+      return { ...provided, opacity, transition, color }
+    },
   }
+
+  return (
+    <Select
+      className={classes.container}
+      value={selectedOption}
```
vitto-moz, line 45: Can we align naming to 
```suggestion
      value={selectedValue}
```

packages/web-app/src/components/Dropdown.tsx
```diff
@@ -16,45 +15,52 @@ const styles = (theme: SaladTheme) => ({
...
   onChange?: (value?: any) => void
 }
 
-const customStyles = {
-  singleValue: (provided: any, state: any) => {
-    const opacity = state.isDisabled ? 0.5 : 1
-    const transition = 'opacity 300ms'
-    const color = '#DBF1C1'
+const _Dropdown: FC<Props> = ({ classes, options, value, onChange }) => {
+  const selectedValue = value
```
vitto-moz, line 23: what is the purpose of this variable set

packages/web-app/src/components/Dropdown.tsx
rners01: added

packages/web-app/src/components/Dropdown.tsx
rners01: refactored

vitto-moz (approved):

Maks19 (approved):

PR #1147, opened 2024-06-12
[WebApp] EarningMachineList - added

vitto-moz:
https://www.notion.so/saladtech/Add-different-colors-checkboxes-for-ID-s-629a4e8147cb49fe8b2fa3c841bb5765?pvs=4
image

### DIFF
--- a/packages/web-app/src/components/Checkbox.tsx
+++ b/packages/web-app/src/components/Checkbox.tsx
@@ -87,6 +87,7 @@ interface Props extends WithStyles<typeof styles> {
   textClassName?: string
   className?: string
   onClick?: (newState: boolean) => void
+  checkedColor?: string
   dark?: boolean
 }
 
@@ -100,8 +101,19 @@ class _Checkbox extends Component<Props> {
   }
 
   public override render(): ReactNode {
-    const { textClassName, disabled, hideCheckbox, className, text, textElement, errorText, checked, dark, classes } =
-      this.props
+    const {
+      textClassName,
+      disabled,
+      hideCheckbox,
+      className,
+      text,
+      textElement,
+      errorText,
+      checked,
+      dark,
+      checkedColor,
+      classes,
+    } = this.props
 
     const hasTextElement = textElement !== undefined
     return (
@@ -116,6 +128,7 @@ class _Checkbox extends Component<Props> {
         >
           {!hideCheckbox && (
             <div
+              style={checked ? { backgroundColor: checkedColor } : undefined}
               className={classnames(classes.checkBox, { [classes.textElementEnabled]: hasTextElement })}
               onClick={this.handleClick}
             >
--- a/packages/web-app/src/modules/earn-views/components/EarningMachineList.tsx
+++ b/packages/web-app/src/modules/earn-views/components/EarningMachineList.tsx
@@ -1,0 +1,93 @@
+import { Text } from '@saladtechnologies/garden-components'
+import type CSS from 'csstype'
+import { useState, type FC } from 'react'
+import Scrollbars from 'react-custom-scrollbars-2'
+import type { WithStyles } from 'react-jss'
+import withStyles from 'react-jss'
+import type { SaladTheme } from '../../../SaladTheme'
+import { Checkbox } from '../../../components'
+
+const styles: (theme: SaladTheme) => Record<string, CSS.Properties> = (theme: SaladTheme) => ({
+  container: {
+    display: 'flex',
+    flexDirection: 'column',
+    position: 'relative',
+    height: '200px',
+    width: '200px',
+    overflow: 'hidden',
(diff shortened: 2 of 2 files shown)


packages/web-app/src/modules/earn-views/components/EarningMachineList.tsx
```diff
@@ -0,0 +1,93 @@
...
+      [machineId]: {
+        ...machineOptions[machineId],
+        isChecked: !machineOptions[machineId]?.isChecked,
+      },
+    } as MachineOptions)
+  }
+
+  return (
+    <div className={classes.container}>
+      <Scrollbars>
```
Maks19, line 77: What do you think about making it more visible with another color?

vitto-moz: look the same as we have requested in the Figma Design file to me. We can discuss this on demo though

packages/web-app/src/modules/earn-views/components/EarningMachineList.tsx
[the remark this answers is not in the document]
vitto-moz: Added 
image

packages/web-app/src/modules/earn-views/components/EarningMachineList.tsx
```diff
@@ -0,0 +1,119 @@
...
+  const totalMachinesAmount = machineOptionsList.length
+  const checkedMachinesAmount = machineOptionsList.filter((machineOption) => machineOption.isChecked).length
+
+  const title = `Machines (${checkedMachinesAmount}/${totalMachinesAmount})`
+
+  return (
+    <div className={classes.container}>
+      <div className={classes.header}>
+        <Text variant="baseM">{title}</Text>
+        <a href="#">
```
Maks19, line 97: You can use SmartLink here

packages/web-app/src/modules/earn-views/components/EarningMachineList.tsx
```diff
@@ -0,0 +1,119 @@
...
+  const [machineOptions, setMachineOptions] = useState<MachineOptions>(mockedMachines)
+
+  const handleMachineOptionClick = (machineId: string) => {
+    setMachineOptions({
+      ...machineOptions,
+      [machineId]: {
+        ...machineOptions[machineId],
+        isChecked: !machineOptions[machineId]?.isChecked,
+      },
+    } as MachineOptions)
```
Maks19, line 84:
```suggestion
    setMachineOptions(prevMachineOptionsState => ({
      ...prevMachineOptionsState,
      [machineId]: {
        ...prevMachineOptionsState[machineId],
        isChecked: !prevMachineOptionsState[machineId]?.isChecked,
      },
    }) as MachineOptions)
```

It makes clear that the new state depends on the previous state.

packages/web-app/src/components/Checkbox.tsx
```diff
@@ -116,6 +128,7 @@ class _Checkbox extends Component<Props> {
         >
           {!hideCheckbox && (
             <div
+              style={checked ? { backgroundColor: checkedColor } : undefined}
```
Maks19, line 131: Why do you need to use style and not className?

packages/web-app/src/modules/earn-views/components/EarningMachineList.tsx
vitto-moz: it's rather better to use callback because of async nature of `useState`: `setMachineOptions`, agree. 
But I think this code will be removed anyway, as we will add handlers and pass the whole state from outside.

packages/web-app/src/components/Checkbox.tsx
vitto-moz: `checkedColor` is a dynamic value, what is your suggestion, could you share the code, please?

Maks19: Add checkBoxClassName property and put it in className

```suggestion
              className={classnames(classes.checkBox, checkBoxClassName, { [classes.textElementEnabled]: hasTextElement })}
```

packages/web-app/src/modules/earn-views/components/EarningMachineList.tsx
vitto-moz: `<SmartLink/>` has additional analytic tracking functionality, do we need it here?

packages/web-app/src/components/Checkbox.tsx
vitto-moz: looks, more complex to me, and requires more code to implement and use

Maks19: Agree with, let's stay with style

Maks19 (approved):

PR #1148, opened 2024-06-13
[WebApp] EarningChart: Earnings Per Machine Switch - added

vitto-moz:
https://www.notion.so/saladtech/Add-Earnings-Per-Machin-toggle-fb3b74100b94456eb547b1cc47473b17?pvs=4

image

### DIFF
--- a/packages/web-app/src/modules/earn-views/components/EarningChart.tsx
+++ b/packages/web-app/src/modules/earn-views/components/EarningChart.tsx
@@ -1,4 +1,4 @@
-import { DefaultTheme as GardenTheme } from '@saladtechnologies/garden-components'
+import { DefaultTheme as GardenTheme, Switch } from '@saladtechnologies/garden-components'
 import classnames from 'classnames'
 import { uniq } from 'lodash'
 import moment from 'moment'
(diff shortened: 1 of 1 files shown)


packages/web-app/src/modules/earn-views/components/EarningChart.tsx
Maks19, line 1: Let's make this component functional if it's possible

Maks19 (approved):

packages/web-app/src/modules/earn-views/components/EarningChart.tsx
```diff
@@ -92,6 +98,9 @@ const styles = (theme: SaladTheme) => ({
   tickMobile: {
     display: 'block',
   },
+  earningPerMachineSwitchWrapper: {
+    marginLeft: 32,
```
Maks19, line 102: Let's decide about approach how we want' to use it 32 or "32px"

vitto-moz: ideally we have to use `CSS.Properties` type, that requires use string format `32px`
```
const styles: (theme: SaladTheme) => Record<string, CSS.Properties> = (theme: SaladTheme) => ({
```
 
 I was not able to apply it here fast, because it required significant refactor of the existing styling due to types discrepancy, so I've decided to just follow the same approach we have in this component.

packages/web-app/src/modules/earn-views/components/EarningChart.tsx
vitto-moz: what the benefit of refactoring this huge component now?

Maks19: Not refactoring it later :)
I propose to make a refactor in Web-App when we need to work with the old component

PR #1159, opened 2024-07-16
[WebApp] disabling Referrals Page based on Support Tool settings - added

vitto-moz:
https://www.notion.so/saladtech/WebApp-When-user-referral-system-off-in-the-web-app-disable-the-Referrals-section-5d8b019e4493457285832f419d8cb28e?pvs=4


Maks19:
I think you should exclude path `/account/referrals` if userReferralsEnabled is false

Maks19:
> I think you should exclude path `/account/referrals` if userReferralsEnabled is false

image

Maks19 (approved):

rners01 (approved):

PR #1164, opened 2024-08-02
[WebApp] Backup Codes Page - added

vitto-moz:
https://www.notion.so/saladtech/FE-Web-app-Generate-recovery-codes-Page-UI-only-0d684e6d67ef42f9817aaea893777514?pvs=4
image


Maks19:
I think we need get rid of empty scrollbars 

image

Maks19 (approved):

PR #1169, opened 2024-08-09
[WebApp] Generate default passkey name

vitto-moz:
https://www.notion.so/saladtech/FE-Add-default-passkey-name-generating-functionality-84f0c3e25da14816aa2dcb69b08ca809?pvs=4


vitto-moz:
should be draft

PR #1170, opened 2024-08-09
[WebApp] Generate default passkey name

vitto-moz:
https://www.notion.so/saladtech/FE-Add-default-passkey-name-generating-functionality-84f0c3e25da14816aa2dcb69b08ca809?pvs=4

### DIFF
--- a/packages/web-app/src/modules/passkey-setup/components/utils.ts
+++ b/packages/web-app/src/modules/passkey-setup/components/utils.ts
@@ -1,8 +1,5 @@
 const getBrowserName = () => {
-  const opera = navigator.userAgent.indexOf('OPR') !== -1
-  console.log('opera ===> ', opera)
   if (navigator.userAgent.indexOf('Opera') !== -1 || navigator.userAgent.indexOf('OPR') !== -1) {
-    console.log('test ===> ')
     return 'Opera'
     // @ts-ignore
   } else if (navigator.userAgent.indexOf('Edg') !== -1) {
@@ -18,7 +15,7 @@ const getBrowserName = () => {
     //IF IE > 10
     return 'IE'
   } else {
-    return 'Unknown'
+    return 'Unknown Browser'
   }
 }
 
@@ -38,7 +35,7 @@ const getOsName = () => {
   } else if (/iPhone|iPad|iPod/.test(userAgent)) {
     os = 'iOS'
   } else {
-    os = 'Unknown'
+    os = 'Unknown OS'
   }
 
   return os
--- a/packages/web-app/src/modules/passkey-setup/components/utils.ts
+++ b/packages/web-app/src/modules/passkey-setup/components/utils.ts
@@ -1,0 +1,49 @@
+const getBrowserName = () => {
+  const opera = navigator.userAgent.indexOf('OPR') !== -1
+  console.log('opera ===> ', opera)
+  if (navigator.userAgent.indexOf('Opera') !== -1 || navigator.userAgent.indexOf('OPR') !== -1) {
+    console.log('test ===> ')
+    return 'Opera'
+    // @ts-ignore
+  } else if (navigator.userAgent.indexOf('Edg') !== -1) {
+    return 'Edge'
+  } else if (navigator.userAgent.indexOf('Chrome') !== -1) {
+    return 'Chrome'
+  } else if (navigator.userAgent.indexOf('Safari') !== -1) {
+    return 'Safari'
+  } else if (navigator.userAgent.indexOf('Firefox') !== -1) {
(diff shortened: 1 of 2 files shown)


packages/web-app/src/modules/passkey-setup/components/utils.ts
```diff
@@ -0,0 +1,49 @@
+const getBrowserName = () => {
+  const opera = navigator.userAgent.indexOf('OPR') !== -1
+  console.log('opera ===> ', opera)
```
Maks19, line 3: I think we can remove it now

packages/web-app/src/modules/passkey-setup/components/utils.ts
```diff
@@ -0,0 +1,49 @@
...
+  if (navigator.userAgent.indexOf('Opera') !== -1 || navigator.userAgent.indexOf('OPR') !== -1) {
+    console.log('test ===> ')
```
Maks19, line 5: I think we can remove it now

packages/web-app/src/modules/passkey-setup/components/utils.ts
```diff
@@ -0,0 +1,49 @@
+const getBrowserName = () => {
+  const opera = navigator.userAgent.indexOf('OPR') !== -1
```
Maks19, line 2: Why do you declare opera here, and not in if function?

packages/web-app/src/modules/passkey-setup/components/utils.ts
```diff
@@ -0,0 +1,49 @@
...
+  console.log('opera ===> ', opera)
+  if (navigator.userAgent.indexOf('Opera') !== -1 || navigator.userAgent.indexOf('OPR') !== -1) {
+    console.log('test ===> ')
+    return 'Opera'
+    // @ts-ignore
+  } else if (navigator.userAgent.indexOf('Edg') !== -1) {
```
Maks19, line 8: It looks like switch case construction. Why do you use if else?

packages/web-app/src/modules/passkey-setup/components/utils.ts
```diff
@@ -0,0 +1,49 @@
...
+  } else if (navigator.userAgent.indexOf('Safari') !== -1) {
+    return 'Safari'
+  } else if (navigator.userAgent.indexOf('Firefox') !== -1) {
+    return 'Firefox'
+    // @ts-ignore
+  } else if (navigator.userAgent.indexOf('MSIE') !== -1 || !!document.documentMode === true) {
+    //IF IE > 10
+    return 'IE'
+  } else {
+    return 'Unknown'
```
Maks19, line 21: What do you think about using more meaningful default names if we can't get browser and os?
For example, 'Default Passkey', as we had previously.

Maks19 (changes requested):

packages/web-app/src/modules/passkey-setup/components/utils.ts
vitto-moz: Could you suggest desired refactor code?

packages/web-app/src/modules/passkey-setup/components/utils.ts
vitto-moz: changed to "Unknown Browser"

packages/web-app/src/modules/passkey-setup/components/utils.ts
vitto-moz: this line is not necessary at all 
removed

packages/web-app/src/modules/passkey-setup/components/utils.ts
Maks19:
```
switch (true) {
  case userAgent.indexOf('Opera') !== -1 || userAgent.indexOf('OPR') !== -1:
    return 'Opera';

  case userAgent.indexOf('Edg') !== -1:
    return 'Edge';

  case userAgent.indexOf('Chrome') !== -1:
    return 'Chrome';

  case userAgent.indexOf('Safari') !== -1:
    return 'Safari';

  case userAgent.indexOf('Firefox') !== -1:
    return 'Firefox';

  case userAgent.indexOf('MSIE') !== -1 || !!document.documentMode === true:
    return 'IE';

  default:
    return 'Unknown Browser';
}
```

Maks19: Any reason to use if else instead of switch(true)?

vitto-moz: refactored to switch

Maks19 (approved):

PR #1179, opened 2024-08-16
[WebApp] Passkeys mobile pages

vitto-moz:
https://www.notion.so/saladtech/FE-Passkeys-Pages-mobile-layots-34cfa6a0e0334e2b93346fd6f377d5aa?pvs=4

### DIFF
--- a/packages/web-app/src/components/TextField.tsx
+++ b/packages/web-app/src/components/TextField.tsx
@@ -9,16 +9,18 @@ import type { SaladTheme } from '../SaladTheme'
 const styles = (theme: SaladTheme) => ({
   container: {
     display: 'inline-block',
+    width: '100%',
   },
   text: {
     border: `1px solid ${theme.lightGreen}`,
-    padding: '.5rem',
+    padding: '24px 16px 24px 16px',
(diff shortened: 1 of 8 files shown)


packages/web-app/src/components/TextField.tsx
```diff
@@ -61,12 +64,14 @@ interface Props extends WithStyles<typeof styles> {
 
 class _TextField extends Component<Props> {
```
Maks19, line 65: Let's make functional components

packages/web-app/src/modules/passkey-delete/components/PasskeyDeletePage.tsx
```diff
@@ -28,9 +28,18 @@ const styles: (theme: SaladTheme) => Record<string, CSS.Properties> = (theme: Sa
     justifyContent: 'left',
     marginTop: '82px',
     marginLeft: '40px',
+    '@media (max-width: 720px)': {
```
Maks19, line 31: I think we use 812px as mobile view

Maks19 (approved):

PR #1200, opened 2024-09-06
[WebApp] Fix proton pass password manager issue

Maks19:
**Description:** Fix proton pass password manager issue using not null value for displayName

### DIFF
--- a/packages/web-app/src/modules/passkey-setup/utils.ts
+++ b/packages/web-app/src/modules/passkey-setup/utils.ts
@@ -103,7 +103,6 @@ interface CredentialOptions {
   user: {
     name: string
     id: string
-    displayName: string
   }
   challenge: string
   pubKeyCredParams: Array<PublicKeyCredentialParameters>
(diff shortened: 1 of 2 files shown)


vitto-moz (dismissed):

packages/web-app/src/modules/passkey-setup/utils.ts
```diff
@@ -103,7 +103,7 @@ interface CredentialOptions {
   user: {
     name: string
     id: string
-    displayName: string
+    displayName: string | null
```
vitto-moz, line 106: shouldn't we use only `string` here?

vitto-moz (approved):

packages/web-app/src/modules/passkey-setup/utils.ts
Maks19: API returns us null by default
Only if we pass something to `/options` endpoint it return string

PR #1212, opened 2024-10-15
[SALAD-22312] WebApp - Add og meta tags for store and rewards

Maks19:
**Description:** Add og meta tags for store and rewards


packages/web-app/src/modules/storefront-views/pages/StorefrontPage.tsx
```diff
@@ -31,6 +32,18 @@ interface Props extends WithStyles<typeof styles> {
...
+
+  const ogMetaTags = {
+    url: href,
+    title: 'Salad - Official Store',
+    description:
+      'Salad helps you earn your way to new Steam Games, Discord Nitro, gift cards, and more from the Salad Storefront. You can even send Salad Balance to PayPal and redeem digital Visa and Mastercard that can be used worldwide!',
+    image: `${origin}/og-official-store.png`,
+    'image:alt': 'Salad - Official Store',
+  }
+  useAddOgMetaTags(ogMetaTags)
```
vitto-moz, line 45: Does this hook usage on `StorefrontPage` cover all pages? for example `https://salad.com/earn/summary`

packages/web-app/src/hooks/useOgMetaTags.tsx
```diff
@@ -0,0 +1,19 @@
...
+        ogMetaTag.setAttribute('content', ogMetaTagContent)
+      } else {
+        const newOgMetaTag = document.createElement('meta')
+        newOgMetaTag.setAttribute('property', `og:${ogMetaTagPropertyName}`)
+        newOgMetaTag.setAttribute('content', ogMetaTagContent)
+        document.head.appendChild(newOgMetaTag)
+      }
+    })
+  }, [ogMetaTagsRecord])
+}
```
vitto-moz, line 19: 1. Let's add some default meta tags values
2. let's rename `ogMetaTagsRecord` -> `ogMetaTags`  to avoid including data type in the variable name 

```suggestion
import { useEffect } from 'react'

const defaultOgMetaTags = {
  url: window.location.href,
  image: `${window.location.origin}/og-official-store.png`,
}

export const useOgMetaTags = (ogMetaTags: Record<string, string>) => {
  useEffect(() => {
    Object.keys({ ...ogMetaTags, ...defaultOgMetaTags }).forEach((ogMetaTagPropertyName) => {
      const ogMetaTag = document.querySelector(`meta[property='og:${ogMetaTagPropertyName}']`)
      const ogMetaTagContent = ogMetaTags[ogMetaTagPropertyName] as string

      if (ogMetaTag) {
        ogMetaTag.setAttribute('content', ogMetaTagContent)
      } else {
        const newOgMetaTag = document.createElement('meta')
        newOgMetaTag.setAttribute('property', `og:${ogMetaTagPropertyName}`)
        newOgMetaTag.setAttribute('content', ogMetaTagContent)
        document.head.appendChild(newOgMetaTag)
      }
    })
  }, [ogMetaTags])
}

```

packages/web-app/src/modules/storefront-views/pages/StorefrontPage.tsx
Maks19: We should cover only `/store`

packages/web-app/src/hooks/useOgMetaTags.tsx
Maks19: I agree with URL but I think image shouldn't be included in default tags

vitto-moz: Displaying the default image in case there wasn't some specific passed - sounds correct to me, any reason to avoid this behavior?

Maks19: I thought you proposed changing every image to the default. Displaying the default image in case there wasn't some specific passed is a good idea. Changed

vitto-moz (approved):

PR #1226, opened 2024-11-27
[SALAD-23099] WebApp: /profile 401: clean localStorage

vitto-moz:


packages/web-app/src/modules/auth/AuthStore.tsx
Maks19, line 1: We clean localStorage in case if /profile 401 but we don't do it if user logs out with `Logout` button. Is it expected? Or should we add localStorage.clear() for logout action?

vitto-moz: the initial purpose to clean storage is for deleted users, so for now we can leave it on logout.

Maks19 (approved):

PR #1236, opened 2024-12-04
[SALAD-23182] WebApp: DemandMonitorPage: onLoginClick - tracking added

vitto-moz:


rners01 (approved):

rners01:
Do you think we need to differentiate what journey leaded to this login for analytics?
Like giving additional info or changing event name to include path or smth

vitto-moz:
> Do you think we need to differentiate what journey leaded to this login for analytics?
> Like giving additional info or changing event name to include path or smth

@rners01, I've focused this part during discussions, but will rise this question one more time on demo.

Maks19 (approved):

PR #1240, opened 2024-12-09
[SALAD-23183] WebApp - Redirect to Demand Alerts Page on successful login

Maks19:
**Description:** Redirect to Demand Alerts Page on successful login

### DIFF
--- a/packages/web-app/src/modules/auth/AuthStore.tsx
+++ b/packages/web-app/src/modules/auth/AuthStore.tsx
@@ -36,12 +36,12 @@ export class AuthStore {
   }
 
   @action
-  public login = async (): Promise<void> => {
+  public login = async (redirect_uri?: string): Promise<void> => {
     // Don't do anything if we are already logged in
     try {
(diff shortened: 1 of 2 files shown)


packages/web-app/src/modules/auth/AuthStore.tsx
```diff
@@ -36,12 +36,12 @@ export class AuthStore {
   }
 
   @action
-  public login = async (): Promise<void> => {
+  public login = async (redirect_uri?: string): Promise<void> => {
```
vitto-moz, line 39:
```suggestion
  public login = async (redirectUri?: string): Promise<void> => {
```

vitto-moz (changes requested):

packages/web-app/src/modules/auth/AuthStore.tsx
Maks19: Updated

vitto-moz (approved):

PR #1242, opened 2024-12-10
[SALAD-23117] WebApp - Add marketing touchpoint tracking

Maks19:
**Description:** Add marketing touchpoint tracking

### DIFF
--- a/packages/web-app/src/App.tsx
+++ b/packages/web-app/src/App.tsx
@@ -19,7 +19,6 @@ import { NovuNotificationBanner } from './modules/notifications-views/components
 import { Routes } from './Routes'
 import type { SaladTheme } from './SaladTheme'
 import type { RootStore } from './Store'
-import { getCookie } from './utils'
 
 const styles = (theme: SaladTheme) => ({
   mainWindow: {
@@ -109,7 +108,6 @@ const searchConfig = {
 interface AppProps extends WithStyles<typeof styles> {
   isAuthenticated: boolean
   setErrorBoundary: (errorBoundary: UseErrorBoundaryApi<Error>) => void
-  trackMarketingTouchpoint: (marketingTouchpointTimestamp: string, utmTags: Record<string, string>) => void
   withInstallReminder: boolean
   novuSignature: string
   history: History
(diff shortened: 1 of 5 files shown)


packages/web-app/src/App.tsx
```diff
@@ -131,6 +134,27 @@ export const _App = ({
     setErrorBoundary(errorBoundary)
   }, [setErrorBoundary, errorBoundary])
 
+  useEffect(() => {
```
vitto-moz, line 137: let's move all this stuff just in `AnalyticsStore.tsx`
there is  no need to keep it in component

Maks19: I have the conditional when to call `trackMarketingTouchpoint`. If I move all this logic to `trackMarketingTouchpoint` in `AnalyticsStore.tsx` I’d call this method every time, but it won’t always trigger Mixpanel tracking.

vitto-moz:
the implementation with mobx `when`

```

  @observable
  private mixpanelInitialized = false

  constructor(private readonly auth: AuthStore) {
    hotjar.initialize({ id: 2225817, sv: 6 })

    const token = config.mixpanelToken
    if (!token) {
      return
    }

    mixpanel.init(token, {
      api_host: `${config.apiBaseUrl}/api/v2/mixpanel`,
      ignore_dnt: true,
      secure_cookie: true,
      loaded: () => {
        runInAction(() => {
          this.mixpanelInitialized = true
        })
      },
    })

    this.trackMarketingTouchpoint()
  }

  /** Tracks marketing touchpoint data when available from cookies and local storage. */
  public trackMarketingTouchpoint = () => {
    when(
      () => !!this.auth.isAuthenticated && this.mixpanelInitialized,
      () => {
        console.log('when ===> ')
        const marketingTouchpointTimestamp = localStorage.getItem('marketingTouchpointTimestamp')
        const utmTags = getUtmTagsFromCookie()
        console.log('utmTags ===> ', utmTags)

        if (marketingTouchpointTimestamp && Object.keys(utmTags).length > 0) {
          console.log('test ===> ', marketingTouchpointTimestamp)
          localStorage.removeItem('marketingTouchpointTimestamp')
          this.track('Marketing Touchpoint', { OriginalTimestamp: marketingTouchpointTimestamp, ...utmTags })
        }
      },
    )
  }

vitto-moz (approved):

PR #1257, opened 2025-01-17
WebApp: load styles for mailchimp form ahead of time / second attempt mailchi…

vitto-moz:
- load styles for mailchimp form ahead of time 
- issue fixed: second attempt mailchimp form usage leads to wrong in-demand device selected


vitto-moz:
wrong branch

PR #1263, opened 2025-01-24
[SALAD-23959] WebApp: Reusable Table component - added

vitto-moz:
image

### DIFF
--- a/packages/web-app/src/components/Table/types.ts
+++ b/packages/web-app/src/components/Table/types.ts
@@ -1,0 +1,8 @@
+type TableRowItem = string | number | JSX.Element
+
+type TableRow = Array<TableRowItem>
+
+export interface TableContent {
+  titles: Array<JSX.Element | string>
+  rows: Array<TableRow>
+}
--- a/packages/web-app/src/components/Table/types.ts
+++ b/packages/web-app/src/components/Table/types.ts
@@ -1,8 +1,5 @@
-type TableRowItem = string | number | JSX.Element
+export type TableRowItem = string | number | JSX.Element
 
-type TableRow = Array<TableRowItem>
+export type TableRow = Array<TableRowItem>
 
-export interface TableContent {
-  titles: Array<JSX.Element | string>
-  rows: Array<TableRow>
-}
+export type TableTitles = Array<JSX.Element | string>
--- a/packages/web-app/src/modules/earn-views/components/AllMachines/AllMachines.tsx
+++ b/packages/web-app/src/modules/earn-views/components/AllMachines/AllMachines.tsx
@@ -4,6 +4,7 @@ import { Checkbox, Text } from '@saladtechnologies/garden-components'
 import classNames from 'classnames'
 import type CSS from 'csstype'
 import { DateTime } from 'luxon'
+import { useState } from 'react'
 import type { WithStyles } from 'react-jss'
 import withStyles from 'react-jss'
 import { Table } from '../../../../components/Table'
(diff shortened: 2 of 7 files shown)


packages/web-app/src/components/Table/types.ts
```diff
@@ -0,0 +1,8 @@
+type TableRowItem = string | number | JSX.Element
+
+type TableRow = Array<TableRowItem>
+
+export interface TableContent {
+  titles: Array<JSX.Element | string>
+  rows: Array<TableRow>
```
rners01, line 7: array of array?

packages/web-app/src/modules/earn-views/components/AllMachines/AllMachines.tsx
```diff
@@ -0,0 +1,145 @@
...
+        <Text variant="baseXS">Warnings</Text>
+      </div>,
+    ]
+  }
+
+  const getRows = (): TableContent['rows'] => {
+    return generatedMockedMachines
+      .map((machineRow) => {
+        return {
+          checkboxes: (
```
rners01, line 100: should it be singular?
`checkbox`

do we have it typed?

packages/web-app/src/components/Table/types.ts
vitto-moz: yes

packages/web-app/src/modules/earn-views/components/AllMachines/AllMachines.tsx
vitto-moz: yes, you are right - it should be singular. We don't need to have it typed, it's just made to mark the columns and make easier to navigate among them, but in fact the keys are not used, you see that the following step it's just transformation of this row object to array: 
```
.map((machineRow) =>
        Object.values(machineRow).map((machineRowItem) => {
          return <div className={classes.tableCell}>{machineRowItem}</div>
        }),

File: packages/web-app/src/modules/earn-views/components/AllMachines/AllMachines.tsx [hunk 3]
```diff
@@ -0,0 +1,166 @@
...
+})
+
+interface Props extends WithStyles<typeof styles> {}
+
+const _AllMachines = ({ classes }: Props) => {
+  const [selectedMachineIds, setSelectedMachineIds] = useState<Record<string, boolean>>(() =>
+    generatedMockedMachines.reduce((aggregatedSelectedMachineIds, machine) => {
+      return { ...aggregatedSelectedMachineIds, [machine.id]: false }
+    }, {}),
+  )
```
Maks19, line 89:
```suggestion
  const [selectedMachineIds, setSelectedMachineIds] = useState<Record<string, boolean>>(() =>
    Object.fromEntries(generatedMockedMachines.map(machine => [machine.id, false])),
  )
```

Maks19: Your variant is good, but it can be hard to read it

packages/web-app/src/modules/earn-views/components/AllMachines/AllMachines.tsx
```diff
@@ -0,0 +1,166 @@
...
+
+  const getTitles = () => {
```
Maks19, line 91:
```suggestion
  const getTitles = useMemo(() => {
```

Maks19: It will prevent unnecessary re-renders each time the checkbox is checked, for example.

packages/web-app/src/modules/earn-views/pages/EarningSummaryPage.tsx
```diff
@@ -88,6 +89,7 @@ const _EarningSummaryPage: FC<Props> = ({
           viewLast7Days={viewLast7Days}
           viewLast30Days={viewLast30Days}
         />
+        <AllMachines />
```
Maks19, line 92: Let's hide it for now

packages/web-app/src/modules/earn-views/components/AllMachines/AllMachines.tsx
vitto-moz: I guess you refer to the render of `<Table/>` component, but: 
1.  `tiles` is not the only prop that `Table` component depends on 
2. Checkboxes are not the part of titles

Maks19 (approved):

PR #1272, opened 2025-01-31
[SALAD-23967] WebApp: Earning Table - added

vitto-moz:


packages/web-app/src/modules/earn-views/components/EarningTable/EarningTable.tsx
```diff
@@ -0,0 +1,182 @@
...
+    ]
+  }
+
+  if (!machines) {
+    return
+  }
+
+  const getRows = (): Array<TableRow> => {
+    return machines
+      .filter((machine) => machine.machine_id)
```
rners01, line 136: why do we need this filter?

packages/web-app/src/modules/earn-views/components/EarningTable/EarningTable.tsx
```diff
@@ -0,0 +1,182 @@
...
+  const {
+    lowestItemNumberOnPage,
+    highestItemNumberOnPage,
+    currentPageNumber,
+    itemsPerPageAmount,
+    setCurrentPageNumber,
+  } = usePagination()
+
+  console.log('daysShowing ===> ', daysShowing)
+  console.log('earningsPerMachine ===> ', earningsPerMachine)
```
rners01, line 113: remove?

vitto-moz: we have temporary keep them to avoid ts errors. Current implementation is partial and we will need this data in future - so no need to remove these props and related code

PR #1279, opened 2025-02-14
[SALAD-24280] WebApp: Dropdown component

vitto-moz:
image

### DIFF
--- a/packages/web-app/src/components/Dropdown/Dropdown.tsx
+++ b/packages/web-app/src/components/Dropdown/Dropdown.tsx
@@ -201,19 +201,17 @@ const _Dropdown = <TOptionValue,>({
         {toggleContent}
       </div>
       {isDropdownShown && (
-        <>
-          <div
-            className={classNames(
-              classes.dropdownWrap,
-              animationType === AnimationType.appearing ? classes.shown : classes.hidden,
-              wrapClassname,
-            )}
-            ref={dropdownRef}
-            onAnimationEnd={onAnimationEnd}
-          >
-            <div className={classes.dropdownContent}>{getDropdownContent()}</div>
-          </div>
-        </>
+        <div
+          className={classNames(
+            classes.dropdownWrap,
+            animationType === AnimationType.appearing ? classes.shown : classes.hidden,
+            wrapClassname,
+          )}
+          ref={dropdownRef}
+          onAnimationEnd={onAnimationEnd}
+        >
+          <div className={classes.dropdownContent}>{getDropdownContent()}</div>
+        </div>
       )}
     </>
   )
--- a/packages/web-app/src/components/Dropdown/Dropdown.tsx
+++ b/packages/web-app/src/components/Dropdown/Dropdown.tsx
@@ -129,7 +129,7 @@ function _Dropdown<TOptionValue = void>({
   }, [isDropdownShownExternal])
 
   useEffect(() => {
-    function handleClickOutside(event: MouseEvent) {
+    const handleClickOutside = (event: MouseEvent) => {
       if (event.target !== toggleDropdownRef.current)
         if (dropdownRef.current && !dropdownRef.current.contains(event.target as Node)) {
           setAnimationType(AnimationType.disappearing)
--- a/packages/web-app/src/components/Dropdown/Dropdown.tsx
+++ b/packages/web-app/src/components/Dropdown/Dropdown.tsx
@@ -91,7 +91,7 @@ export interface Props<TOptionValue> extends WithStyles<typeof styles> {
   optionKey?: keyof DropdownOption
 }
 
-function _Dropdown<TOptionValue = void>({
+const _Dropdown = <TOptionValue,>({
   options,
   content,
   toggleContent,
@@ -100,7 +100,7 @@ function _Dropdown<TOptionValue = void>({
   setDropdownShown: setDropdownShownExternal,
   optionKey,
   classes,
-}: Props<TOptionValue>): React.ReactElement {
+}: Props<TOptionValue>): React.ReactElement => {
   if (
     (setDropdownShownExternal === undefined && isDropdownShownExternal) ||
     (setDropdownShownExternal && isDropdownShownExternal === undefined)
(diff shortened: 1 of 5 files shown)


packages/web-app/src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,216 @@
...
+  options?: DropdownOption<TOptionValue>[]
+  content?: JSX.Element
+  toggleContent?: JSX.Element
+  wrapClassname?: string
+  isDropdownShown?: boolean
+  setDropdownShown?: (value: boolean) => void
+  optionKey?: keyof DropdownOption
+}
+
+function _Dropdown<TOptionValue = void>({
```
Maks19, line 94: Let's use 
const _Dropdown

packages/web-app/src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,216 @@
...
+    throw new Error(
+      'Dropdown: You should provide both "isDropdownShown" and "setDropdownShown" props in case of external state usage',
+    )
+  }
+
+  if ((optionKey === undefined && options) || (optionKey && options === undefined)) {
+    throw new Error(
+      'Dropdown: You should provide both "options" and "optionKey" props in case passing options instead of content',
+    )
+  }
```
Maks19, line 117: Why do you need to throw Error inside the component?

packages/web-app/src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,216 @@
...
+  setDropdownShown?: (value: boolean) => void
+  optionKey?: keyof DropdownOption
+}
+
+function _Dropdown<TOptionValue = void>({
+  options,
+  content,
+  toggleContent,
+  wrapClassname,
+  isDropdownShown: isDropdownShownExternal,
```
Maks19, line 99: Where do you use isDropdownShown prop?

packages/web-app/src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,216 @@
...
+
+  useEffect(() => {
+    if (isDropdownShownExternal !== undefined) {
+      setAnimationType(isDropdownShownExternal ? AnimationType.appearing : AnimationType.disappearing)
+      setDropdownShown(isDropdownShownExternal)
+    }
+  }, [isDropdownShownExternal])
+
+  useEffect(() => {
+    function handleClickOutside(event: MouseEvent) {
```
Maks19, line 132:
```suggestion
    const handleClickOutside = (event: MouseEvent) => {
```

packages/web-app/src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,216 @@
...
+              wrapClassname,
+            )}
+            ref={dropdownRef}
+            onAnimationEnd={onAnimationEnd}
+          >
+            <div className={classes.dropdownContent}>
+              {content
+                ? content
+                : options && optionKey
+                ? options
```
Maks19, line 186: Let's avoid using multiple ternary operators

packages/web-app/src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,216 @@
...
+                    .filter((option) => option.shown === undefined || option.shown)
+                    .map((option) => {
+                      if (typeof option[optionKey] !== 'string') {
+                        throw new Error('Dropdown: optionKey props should point the unique string value to use as key')
+                      }
```
Maks19, line 191: Why do we need to throw error?

packages/web-app/src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,216 @@
...
+    option.handler?.(option)
+  }
+
+  return (
+    <>
+      <div className={classes.toggleDropdown} onClick={toggleDropdown}>
+        {toggleContent}
+      </div>
+      {isDropdownShown && (
+        <>
```
Maks19, line 172:
```suggestion

```

Maks19: It can be removed

Maks19 (changes requested):

packages/web-app/src/components/Dropdown/Dropdown.tsx
vitto-moz: As this component was made as multi-purpose there are variations how we can use it: 

Internal and external state options: 
- work without external state (only local state to keep dropdown shown or not)
- work with external state (show the dropdown based on outer state passed with prop)

Content displayed in the dropdown: 
- display options list (passing options)
- display content (just show the any content passed as a component into `content` prop) 

Every usage requires its set of props passed, so the errors control the right set of props has been passed.

packages/web-app/src/components/Dropdown/Dropdown.tsx
vitto-moz: as Dropdown component renders options as the list, which can be dynamically changed in a runtime - we have to set some unique key for every item component of the list. So it's important that the `optionKey` is present and can be use as the list component key

packages/web-app/src/modules/earn-views/pages/EarningSummaryPage.tsx
```diff
@@ -50,22 +51,28 @@ const _EarningSummaryPage: FC<Props> = ({
...
   trackEarnPageFAQLinkClicked,
   trackEarnPageViewed,
   viewLast24Hours,
   viewLast7Days,
   viewLast30Days,
 }) => {
-  const [selectedMachineId, setSelectedMachineId] = useState<string | null>(null)
+  const [shownModalMachineId, setShownModalMachineId] = useState<string | null>(null)
+  const [selectedMachineIds, setSelectedMachineIds] = useState<string[]>([])
+  console.log('selectedMachineIds ===> ', selectedMachineIds)
```
Maks19, line 66: We can remove it

vitto-moz: no we cannot - there is only one usage in logs, but we need this state for the following functionality we gonna implement

PR #1314, opened 2025-04-10
[SALAD-25212] WebApp - fix horizontal scrollbar in firefox

Maks19:
**Description:** Fix the horizontal scrollbar problem in Firefox for the Earning History section, reverting to the default scrollbar.

image

### DIFF
--- a/packages/web-app/src/components/Table/Table.tsx
+++ b/packages/web-app/src/components/Table/Table.tsx
@@ -57,6 +57,7 @@ const _Table: FunctionComponent<Props> = ({
   const renderHorizontalTrack: React.ComponentType<any> = ({ style, ...props }) => {
     return (
       <div
+        className="customScrollbarHorizontalTrack"
         style={{
           position: 'absolute',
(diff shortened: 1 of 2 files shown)


packages/web-app/src/components/Table/Table.tsx
```diff
@@ -57,6 +57,7 @@ const _Table: FunctionComponent<Props> = ({
   const renderHorizontalTrack: React.ComponentType<any> = ({ style, ...props }) => {
     return (
       <div
+        className="customScrollbarHorizontalTrack"
```
vitto-moz, line 60: can this be covered with `react-jss` instead of adding global styling?

Maks19: No, we cannot replace `@-moz-document` rule using `react-jss`. `@-moz-document` is a non-standard, Firefox-specific at-rule that is only valid in actual CSS files.

vitto-moz (approved):

PR #1326, opened 2025-12-05
[Web-App] Remove PayPal Account link from Accounts Page.

AngarosGamer:
Removes the PayPal login field in the account settings, as it is unused.
Display code purposefully added as comments, if needed to be enabled further later.
image
Vs Current:
<REDACTED>


scottcinatl (commented):
Is this something that will be put back soon? If not, then we just delete the code rather than just comment it out.

scottcinatl (changes requested):
There is just one more cleanup in the auth store that still needs to be removed - https://github.com/SaladTechnologies/salad-applications/blob/master/packages/web-app/src/modules/auth/AuthStore.tsx#L10

scottcinatl (approved):

